PR #4, opened 2020-02-06
Correct handling of single end data

FelixMoelder:
This PR addresses an issue (found by @jafors and @dlaehnemann) causing the workflow to break when applied to single end data.
This occurs while merging fastq files as a `read`-wildcard is expected.


workflow/rules/trimming.smk
```diff
@@ -55,7 +55,7 @@ rule merge_fastqs:
     input:
         lambda w: expand("trimmed/{{sample}}-{unit}.{{read}}.fastq.gz", unit=units.loc[w.sample, "unit_name"])
     output:
-        "merged/{sample}.{read,(1|2)}.fastq.gz"
+        "merged/{sample}.{read,(single|1|2)}.fastq.gz"
```
johanneskoester: Actually, the conditional on whether having gz or not is not necessary. gzipped files can be catted like normal files. Further, since both output and input are always gz there wouldn't even be the need to distinguish if it was necessary.

PR #5, opened 2020-02-10
Stream data to gz-file

FelixMoelder:
This PR modifies the rule `dbnsfp_bgzip`:
Previously, data has been written to disk before being compressed to a gzip-file.
This has now been replaced by streaming data directly.


FelixMoelder:
Should we increase the compression level as the output file becomes bigger than the input (27GB vs 25GB)

johanneskoester:
Yes, let us compress as much as possible.

PR #8, opened 2020-03-18
Bug fix and version bump

FelixMoelder:
This PR fixes an issue, identified by @christopher-schroeder, where vcf-content is written into a bcf-file without transforming into binary output.

Additionally, there are some version updates circumventing potential issues in the workflow.


johanneskoester:
There is a conda version conflict in one of the envs.

FelixMoelder:
It seems we have a little issue here.


> There is a conda version conflict in one of the envs.

It seems like we need to go back to bcftools 1.9 for igv-reports as pysam doesn't support v1.10, yet.
But as igv-reports at the end of line that should not be any problem.

FelixMoelder:
> Thanks! Please revert the config file changes, I'd like to keep the homo sapiens example for now.

Oooops. That was unintended.

workflow/rules/annotation/pass2.smk
```diff
@@ -44,7 +44,7 @@ rule annotate_dgidb:
     resources:
         dgidb_requests=1
     shell:
-        "rbt vcf-annotate-dgidb {input} > {output}"
+        "rbt vcf-annotate-dgidb {input} | bcftools view -Ob > {output}"
```
johanneskoester: Sorry, just seen this. I'd rather have rbt print BCF directly. This is how all other rbt commands should be implemented already. Can you change that accordingly?

FelixMoelder: Sure, I will check on that.

FelixMoelder:
The last commit will work as soon as rbt is updated to a new version (https://github.com/rust-bio/rust-bio-tools/pull/46)

PR #10, opened 2020-03-20
Toresults

christopher-schroeder:
Changed the location of the produced files to ./results/

### DIFF
--- a/workflow/rules/annotation/pass2.smk
+++ b/workflow/rules/annotation/pass2.smk
@@ -49,7 +49,7 @@ rule annotate_dgidb:
     resources:
         dgidb_requests=1
     shell:
-        "rbt vcf-annotate-dgidb {input} |  bcftools view --threads {threads} -Ob > {output}"
+        "rbt vcf-annotate-dgidb {input} | bcftools view --threads {threads} -Ob > {output}"
 
 
 if is_activated("annotations/dbnsfp"):
--- a/workflow/rules/annotation/pass2.smk
+++ b/workflow/rules/annotation/pass2.smk
@@ -68,11 +68,9 @@ if is_activated("annotations/dbnsfp"):
         threads: 4
         shell:
             """
-            unzip {input} -d resources/dbnsfp/ &&
-            (zcat resources/dbnsfp/*_variant.chr1.gz |
-            head -n 1 ; zcat resources/dbnsfp/*_variant.chr* |
-            grep -v '^#' ) | bgzip -@ {threads} > {output} &&
-            rm -r resources/dbnsfp
+            (unzip -p {input} "*_variant.chr1.gz" | zcat |
+            head -n 1 ; unzip -p {input} "*_variant.chr*" |
+            zcat | grep -v '^#' ) | bgzip -l9 -@ {threads} > {output}
             """
     
     rule annotate_dbnsfp:
--- a/workflow/rules/annotation/pass2.smk
+++ b/workflow/rules/annotation/pass2.smk
@@ -1,5 +1,6 @@
-annotations = [(e, f) for e, f in config["annotations"]["vcfs"].items() if e != "activate"]
+import os
 
+annotations = [(e, os.path.join("results", f)) for e, f in config["annotations"]["vcfs"].items() if e != "activate"]
 
 def get_annotation_pipes(wildcards, input):
      if annotations:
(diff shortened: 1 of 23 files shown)


workflow/rules/annotation/pass2.smk
```diff
@@ -18,33 +19,37 @@ def get_annotation_vcfs(idx=False):
...
-        "calls/{prefix}.dgidb.bcf"
+        "results/calls/{prefix}.dgidb.bcf"
     params:
     conda:
         "../../envs/annotate_dgidb.yaml"
     resources:
         dgidb_requests=1
     shell:
-        "rbt vcf-annotate-dgidb {input} > {output}"
+        "rbt vcf-annotate-dgidb -g 500 {input} |  bcftools view --threads {threads} -Ob > {output}"
```
FelixMoelder: Explicitly setting genes to 500 is not required as this is already the default in the latest rbt version.

workflow/rules/mapping.smk
```diff
@@ -3,25 +3,25 @@ rule map_reads:
...
+        temp("results/mapped/{sample}.sorted.bam")
     log:
         "logs/bwa_mem/{sample}.log"
     params:
         index=lambda w, input: os.path.splitext(input.idx[0])[0],
         extra=get_read_group,
         sort="samtools",
         sort_order="coordinate"
-    threads: 8
+    threads: 100
```
johanneskoester: Actually, I tried this once after your advice last time, and it seems this does not saturate at all. If I remember correctly, it was still about 8 cores being used most of the time. Might be because we use samtools to sort, instead of sambamba here.

workflow/rules/ref.smk
```diff
@@ -78,9 +78,11 @@ rule tabix_known_variants:
 
 rule bwa_index:
     input:
-        "refs/genome.fasta"
+        "results/refs/genome.fasta"
     output:
-        multiext("refs/genome.fasta", ".amb", ".ann", ".bwt", ".pac", ".sa")
+        multiext("results/refs/genome", ".amb", ".ann", ".bwt", ".pac", ".sa")
```
johanneskoester: We actually have a divergence here. The multiext needs .fasta, and the params can be removed.

workflow/rules/trimming.smk
```diff
@@ -10,25 +12,48 @@ rule get_sra:
...
         # single end local sample
-        return unit["fq1"].values
+        return f"pipe/cutadapt/{unit.sample_name}-{unit.unit_name}.fq1.fastq{ending}"
     else:
         # paired end local sample
-        return unit[["fq1", "fq2"]].values
+        return expand(f"pipe/cutadapt/{unit.sample_name}-{unit.unit_name}.{{read}}.fastq{ending}", read=["fq1","fq2"])
+
+
+def cutadapt_pipe_input(wc):
```
johanneskoester: say `wildcards` here, to improve readability.

workflow/rules/trimming.smk
```diff
@@ -10,25 +12,48 @@ rule get_sra:
...
+    files = list(sorted(glob.glob(units.loc[wc.sample].loc[wc.unit, wc.fq])))
+    #print(wc, units.loc[wc.sample].loc[wc.unit, wc.fq])
+    assert(len(files) > 0)
+    return files
+
+
+rule cutadapt_pipe:
+    threads:
+        0
```
johanneskoester: Always put threads directly above the shell directive. It is most related to that, and less important for reading the step than input and output.

workflow/rules/trimming.smk
```diff
@@ -10,25 +12,48 @@ rule get_sra:
...
 rule cutadapt_pe:
     input:
         get_raw_fastq
     output:
-        fastq1="trimmed/{sample}-{unit}.1.fastq.gz",
-        fastq2="trimmed/{sample}-{unit}.2.fastq.gz",
-        qc="trimmed/{sample}-{unit}.paired.qc.txt"
+        fastq1="results/trimmed/{sample}-{unit}.1.fastq.gz",
+        fastq2="results/trimmed/{sample}-{unit}.2.fastq.gz",
+        qc="results/trimmed/{sample}-{unit}.qc.txt"
```
johanneskoester: Divergence again. Please keep the `.paired.` infix of the qc file and add a `.single` infix to the one in the rule below.

workflow/rules/trimming.smk
```diff
@@ -10,25 +12,48 @@ rule get_sra:
...
+        ending = ""
+
     if pd.isna(unit["fq1"]):
         # SRA sample (always paired-end for now)
         accession = unit["sra"]
         return expand("sra/{accession}_{read}.fastq", accession=accession, read=[1, 2])
     if pd.isna(unit["fq2"]):
         # single end local sample
-        return unit["fq1"].values
+        return f"pipe/cutadapt/{unit.sample_name}-{unit.unit_name}.fq1.fastq{ending}"
```
johanneskoester: Using `f` is in principle nice, but Snakemake still allows python 3.5, which does not have it. I think running this workflow with 3.5 would lead to a syntax error then. Or is it possible to check the version without running into a syntax error with this?

workflow/rules/trimming.smk
```diff
@@ -10,25 +12,48 @@ rule get_sra:
...
+    return files
+
+
+rule cutadapt_pipe:
+    threads:
+        0
+    input:
+        cutadapt_pipe_input
+    output:
+        pipe("pipe/cutadapt/{sample}-{unit}.{fq}.fast{ending}")
```
johanneskoester: Shouldn't this be `.fastq{ending}`?

workflow/rules/mapping.smk
christopher-schroeder: Oh that's weird. Actually this is crucial for me performing WGS analysis. But I replaced this rule with the bwa-samblaster wrapper. I will add a config option to change the cores.

workflow/rules/trimming.smk
christopher-schroeder: no, then the wildcard "ending" would be "", which is not allowed and delivers an error

workflow/rules/ref.smk
christopher-schroeder: Actually I think both is valid and supported.

christopher-schroeder: but I changed it back

workflow/rules/trimming.smk
christopher-schroeder: i changed this using wildcard_constraints

workflow/rules/trimming.smk
christopher-schroeder: ah ok, I will pay attention to this in future PR

workflow/rules/ref.smk
johanneskoester: The thing is, when using a manually set prefix here, you have to ensure that it is always a prefix of the output, even if the output is changed via --default-remote-prefix (in the cloud). Hence, the code becomes more complicated, because you will need a function for the prefix. Since it is easier to just omit the prefix I voted for that.

workflow/rules/trimming.smk
johanneskoester: Great, thanks! THis is much easier to read now.

workflow/rules/mapping.smk
johanneskoester: You don't need a config option. You can (since some recent release) overwrite the threads of any rule at the command line via e.g. `--threads bwa_mem=100`.

PR #13, opened 2020-03-24
Download snpEff database

FelixMoelder:
This PR adds a rule for downloading the snpEff database before annotating.
Previously the download was initiated by each snpEff annotation call.
This could lead to race conditions where each call downloads the database to the same temporary directory which got deleted after the first download finished.

### DIFF
--- a/workflow/rules/annotation/pass1.smk
+++ b/workflow/rules/annotation/pass1.smk
@@ -2,7 +2,7 @@ rule download_snpeff_db:
     output:
         directory("results/refs/snpeff/{ref}")
     params:
-        db_dir=lambda _, output: Path(output).parent.resolve()
+        db_dir=lambda _, output: Path(output[0]).parent.resolve()
     cache: True
     conda:
         "../../envs/snpeff.yaml"
--- a/workflow/rules/annotation/pass1.smk
+++ b/workflow/rules/annotation/pass1.smk
@@ -1,6 +1,18 @@
+rule download_snpeff_db:
+    output:
+        directory("{working_dir}/results/databases/{{ref}}".format(working_dir=os.getcwd()))
+    params:
+        db_dir=os.path.join(os.getcwd(), "results/databases/")
+    cache: True
+    conda:
+        "../../envs/snpeff.yaml"
+    shell:
+        "snpEff download -dataDir {params.db_dir} {wildcards.ref}"
+
 rule snpeff:
     input:
-        "results/calls/{group}.bcf"
+        calls="results/calls/{group}.bcf",
+        db="{working_dir}/results/databases/{build}.{release}".format(working_dir=os.getcwd(), build=config["ref"]["build"], release=config["ref"]["snpeff_release"])
     output:
         calls="results/calls/{group}.annotated.bcf",
         stats="results/snpeff/{group}.html",
@@ -9,6 +21,7 @@ rule snpeff:
         "logs/snpeff/{group}.log"
     params:
         reference="{}.{}".format(config["ref"]["build"], config["ref"]["snpeff_release"]),
-        extra="-Xmx4g"
+        data_dir = os.path.join(os.getcwd(), "results/databases/"),
+        extra="-Xmx4g -nodownload"
     wrapper:
         "0.50.4/bio/snpeff"
--- a/workflow/rules/annotation/pass1.smk
+++ b/workflow/rules/annotation/pass1.smk
@@ -12,7 +12,7 @@ rule download_snpeff_db:
 rule snpeff:
     input:
         calls="results/calls/{group}.bcf",
-        db="results/refs/snpeff/{build}.{release}".format(build=config["ref"]["build"], release=config["ref"]["snpeff_release"])
+        db="results/refs/snpeff/{build}.{snpeff_release}".format(**config["ref"])
     output:
         calls="results/calls/{group}.annotated.bcf",
         stats="results/snpeff/{group}.html",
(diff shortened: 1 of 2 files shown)


workflow/rules/annotation/pass1.smk
```diff
@@ -1,6 +1,18 @@
+rule download_snpeff_db:
+    output:
+        directory("{working_dir}/results/databases/{{ref}}".format(working_dir=os.getcwd()))
```
johanneskoester: The `{working_dir}` is not necessary. Just have a relative path starting with results. It will always be in the working dir. If snpeff expects an absolute path, rather have this translated in params.

workflow/rules/annotation/pass1.smk
```diff
@@ -1,6 +1,18 @@
...
+    params:
+        db_dir=os.path.join(os.getcwd(), "results/databases/")
```
johanneskoester: results/databases is not a portable parameter, because the output path can be altered by flags like --default-remote-prefix. You have to use a function here instead: `lambda _, output: os.path.dirname(output[0])`.

workflow/rules/annotation/pass1.smk
```diff
@@ -1,6 +1,18 @@
...
+    conda:
+        "../../envs/snpeff.yaml"
+    shell:
+        "snpEff download -dataDir {params.db_dir} {wildcards.ref}"
+
 rule snpeff:
     input:
-        "results/calls/{group}.bcf"
+        calls="results/calls/{group}.bcf",
+        db="{working_dir}/results/databases/{build}.{release}".format(working_dir=os.getcwd(), build=config["ref"]["build"], release=config["ref"]["snpeff_release"])
```
johanneskoester: Same here, working_dir not needed.

workflow/rules/annotation/pass1.smk
```diff
@@ -9,6 +21,7 @@ rule snpeff:
         "logs/snpeff/{group}.log"
     params:
         reference="{}.{}".format(config["ref"]["build"], config["ref"]["snpeff_release"]),
-        extra="-Xmx4g"
+        data_dir = os.path.join(os.getcwd(), "results/databases/"),
```
johanneskoester: Does snpeff expect an absolute path? If so, use `lambda _, input: Path(input.db).parent.resolve()`.

johanneskoester: Eventually, you might need to add `str()` around the Path-statement. Depends a bit on how the wrapper uses it.

workflow/rules/annotation/pass1.smk
johanneskoester: Output should go in `results/refs/snpeff`.

FelixMoelder: Right, it is only necessary for the snpeff call and I just translated this syntax to the output.

workflow/rules/annotation/pass1.smk
```diff
@@ -9,6 +21,7 @@ rule snpeff:
         "logs/snpeff/{group}.log"
     params:
         reference="{}.{}".format(config["ref"]["build"], config["ref"]["snpeff_release"]),
-        extra="-Xmx4g"
+        data_dir = lambda _, input: Path(input.db).parent.resolve(),
```
johanneskoester: formatting: no whitespace around `=`.

workflow/rules/annotation/pass1.smk
```diff
@@ -1,6 +1,18 @@
...
+    conda:
+        "../../envs/snpeff.yaml"
+    shell:
+        "snpEff download -dataDir {params.db_dir} {wildcards.ref}"
+
 rule snpeff:
     input:
-        "results/calls/{group}.bcf"
+        calls="results/calls/{group}.bcf",
+        db="results/refs/snpeff/{build}.{release}".format(build=config["ref"]["build"], release=config["ref"]["snpeff_release"])
```
johanneskoester:
```suggestion
        db="results/refs/snpeff/{build}.{snpeff_release}".format(**config["ref"])
```

workflow/rules/annotation/pass1.smk
```diff
@@ -9,6 +21,7 @@ rule snpeff:
         "logs/snpeff/{group}.log"
     params:
         reference="{}.{}".format(config["ref"]["build"], config["ref"]["snpeff_release"]),
-        extra="-Xmx4g"
+        data_dir = lambda _, input: Path(input.db).parent.resolve(),
+        extra="-Xmx4g -nodownload"
```
johanneskoester:
```suggestion
        extra="-Xmx4g -nodownload"
    resources:
        mem_mb=4000
```

workflow/rules/annotation/pass1.smk
```diff
@@ -1,14 +1,29 @@
+rule download_snpeff_db:
+    output:
+        directory("results/refs/snpeff/{ref}")
```
dlaehnemann: Not sure if you have decided on something else here, but generally the idea was to put references and databases in the main directory `resources` to distinguish it from input-data-specific `results`.

The general setup considerations are documented in this fork of the snakemake-workflow cookie-cutter template:
https://github.com/dlaehnemann/cookiecutter-snakemake-workflow/tree/cleaner-main-directory/%7B%7B%20cookiecutter.repo_name%20%7D%7D

And some more discussion is available in the respective PR:
https://github.com/snakemake-workflows/cookiecutter-snakemake-workflow/pull/4

johanneskoester: That is actually a good point. I agree. Let us fix this in a separate follow-up PR. I'll take care of that. Sorry for having forgotten that.

PR #14, opened 2020-03-25
Fix for ambiguous rules

FelixMoelder:
This PR fixes an issue preventing snakemake to build the dag, as the output of rules `create_dbnsfp_tabix_index` and `tabix_known_variants` was undistinguishable.

### DIFF
--- a/workflow/rules/common.smk
+++ b/workflow/rules/common.smk
@@ -170,3 +170,11 @@ def get_annotation_pipes(wildcards, input):
 def get_annotation_vcfs(idx=False):
     fmt = lambda f: f if not idx else f + ".tbi"
     return [fmt(f) for _, f in annotations]
+
+
+def get_tabix_params(wildcards):
+    if wc.format == "vcf":
+        return "-p vcf"
+    if wc.format == "txt":
+        return "-s 1 -b 2 -e 2"
+    return ""
--- a/workflow/rules/common.smk
+++ b/workflow/rules/common.smk
@@ -173,8 +173,8 @@ def get_annotation_vcfs(idx=False):
 
 
 def get_tabix_params(wildcards):
-    if wc.format == "vcf":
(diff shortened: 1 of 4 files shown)


johanneskoester:
Can you merge the two into one rule, and move them into utils.smk?

FelixMoelder:
> Can you merge the two into one rule, and move them into utils.smk?

Done, but I'm not happy about the `get_tabix_params` method as an empty string could be returned. Suggestions?

workflow/rules/common.smk
```diff
@@ -169,4 +169,12 @@ def get_annotation_pipes(wildcards, input):
...
\ No newline at end of file
+    return [fmt(f) for _, f in annotations]
+
+
+def get_tabix_params(wildcards):
+    if wildcards.format == "vcf":
+        return "-p vcf"
+    if wildcards.format == "txt":
+        return "-s 1 -b 2 -e 2"
+    return ""
```
johanneskoester: Raise a value error here, complaining about an invalid format.

workflow/rules/common.smk
```diff
@@ -173,8 +173,10 @@ def get_annotation_vcfs(idx=False):
...
-        return "-s 1 -b 2 -e 2"
-    return ""
\ No newline at end of file
+    try:
+        if wildcards.format == "vcf":
+            return "-p vcf"
+        if wildcards.format == "txt":
+            return "-s 1 -b 2 -e 2"
+    except ValueError:
+        print("Invalid format: {}".format(wildcards.format))
```
johanneskoester:
```suggestion
    if wildcards.format == "vcf":
        return "-p vcf"
    if wildcards.format == "txt":
        return "-s 1 -b 2 -e 2"
    raise ValueError("Invalid format for tabix: {}".format(wildcards.format))
```

Issue #16, opened 2020-03-27
Building genome.dict fails at first try


FelixMoelder:
While running the workflow rule `recalibrate_base_qualities` fails, telling that the genome dictionary in invalid.
To solve this `genome.dict` needs to be deleted and newly created by rule `genome_dict`.
The new `genome.dict` has a bigger file size than the old/first one.
This always happens when running the workflow for the first time.

[get-genome.log](https://github.com/snakemake-workflows/dna-seq-varlociraptor/files/4392308/get-genome.log)
[create_dict_bak.log](https://github.com/snakemake-workflows/dna-seq-varlociraptor/files/4392310/create_dict_bak.log)

Any idea why this happens?

johanneskoester:
How do the two (the invalid and the correct one) genome dicts look like?

FelixMoelder:
The invalid one only has a header without any additional content while the valid one looks like expected.

johanneskoester:
Looks like a weird bug in picard. Let us replace it with samtools: http://www.htslib.org/doc/samtools-dict.html

johanneskoester:
Can you quickly try that out?

FelixMoelder:
Jep, I just started a new run. Could take a moment as the fasta needs to be downloaded again to reproduce the behavior.

jafors:
Just fyi, I tried running only the "get_genome" and "genome_dict" rules, using `snakemake resources/genome.dict --use-conda -j 1 ` and the resulting genome.dict looks perfectly fine.

FelixMoelder:
Strange. When performing "get_genome" before "genome_dict" it does not work for me. Samtools instead seems to work fine.

PR #18, opened 2020-03-27
Replace Picard by samtools

FelixMoelder:
This PR replaces Picardtools by samtools for generating the genome dictionary as Picardtools creates an empty one the call. The issue is further explained [here](https://github.com/snakemake-workflows/dna-seq-varlociraptor/issues/16).

Before merging wait until a test run has finished.


FelixMoelder:
A test run was successful.
So, we can merge this.

PR #19, opened 2020-04-01
Parallel annotation

christopher-schroeder:
I've written a script to perform parallelize executions on vcf. This may not be important for WES or other targeted sequencing data, but it is crucial for me. I have a 220gb compressed vcf.gz database and annotating WGS data. This takes days on a single core.
There might be overhead now for targeted sequencing, but testing shows nearly no decrease in runtime for small data.

I would like to keep to keep the script as a standalone script as it is already on my github. I know that
"(python ../workflow/scripts/parallel_vcf.py --threads {threads} {input.bcf} '{params.pipes}' | bcftools view --threads {threads} -Ob > {output}) 2> {log}"
is probably not the way to go here. Can someone tell me a better way of invoking scripts via shell command?

Best
Christo

### DIFF
--- a/workflow/scripts/parallel_vcf.py
+++ b/workflow/scripts/parallel_vcf.py
@@ -1,0 +1,50 @@
+import os
+import argparse
+from multiprocessing import Pool
+from pysam import VariantFile
+import tempfile
+import sys
+
+parser = argparse.ArgumentParser()
+parser.add_argument("file_vcf")
+parser.add_argument("command")
+parser.add_argument("--stepsize", "-s", type=int, default=1000000)
+parser.add_argument("--threads", "-t", type=int, default=8)
+parser.add_argument("--tmp", default=None, help="Define the base directory for the tmp files")
+args = parser.parse_args()
+tmp_base = args.tmp
+stepsize = args.stepsize
+
+# create regions
+f = VariantFile(args.file_vcf)
+regions = []
+for name, contig in f.header.contigs.items():
+    last = 0
+    for x in list(range(0, contig.length, stepsize)):
+        stop = min(last + stepsize - 1, contig.length)
+        regions.append(f"{name}:{last}-{stop}")
+        last = stop+1
+f.close() 
+
+#command = "SnpSift annotate {ss_args} {args.database_vcf} /dev/stdin"
+command = args.command
+
+with tempfile.TemporaryDirectory(dir=tmp_base) as t:
+    def f(x):
+        #print("start", x, file=sys.stderr)
+        cmd = f"bcftools view -r {x} {args.file_vcf} | bcftools view -t {x} | {command} | bcftools view -Ob > {t}/{x}.bcf"
+        os.system(cmd)
+        #print("done", x, file=sys.stderr)
+        return f"{t}/{x}.bcf"
+
+    with Pool(args.threads) as p:
+        first = True
+        for filename in p.imap(f, regions):
+            print("merge", filename, file=sys.stderr)
+            if first:
+                cmd = f"bcftools view {filename}"
+            else:
+                cmd = f'bcftools view {filename} | grep -v "#"'
+            os.system(cmd)
+            os.remove(filename)
+            first = False
--- a/workflow/scripts/parallel_vcf.py
+++ b/workflow/scripts/parallel_vcf.py
@@ -10,36 +10,36 @@ parser.add_argument("file_vcf")
 parser.add_argument("command")
 parser.add_argument("--stepsize", "-s", type=int, default=1000000)
 parser.add_argument("--threads", "-t", type=int, default=8)
-parser.add_argument("--tmp", default=None, help="Define the base directory for the tmp files")
+parser.add_argument(
+    "--tmp", default=None, help="Define the base directory for the tmp files"
+)
 args = parser.parse_args()
 tmp_base = args.tmp
 stepsize = args.stepsize
 
 # create regions
-f = VariantFile(args.file_vcf)
-regions = []
-for name, contig in f.header.contigs.items():
-    last = 0
-    for x in list(range(0, contig.length, stepsize)):
-        stop = min(last + stepsize - 1, contig.length)
-        regions.append(f"{name}:{last}-{stop}")
-        last = stop+1
-f.close() 
+with VariantFile(args.file_vcf) as f:
+    regions = []
+    for name, contig in f.header.contigs.items():
+        last = 0
+        for x in list(range(0, contig.length, stepsize)):
+            stop = min(last + stepsize - 1, contig.length)
+            regions.append(f"{name}:{last}-{stop}")
+            last = stop + 1
 
-#command = "SnpSift annotate {ss_args} {args.database_vcf} /dev/stdin"
+# command = "SnpSift annotate {ss_args} {args.database_vcf} /dev/stdin"
 command = args.command
 
-with tempfile.TemporaryDirectory(dir=tmp_base) as t:
-    def f(x):
-        #print("start", x, file=sys.stderr)
-        cmd = f"bcftools view -r {x} {args.file_vcf} | bcftools view -t {x} | {command} | bcftools view -Ob > {t}/{x}.bcf"
+with tempfile.TemporaryDirectory(dir=tmp_base) as tmp_dir:
+
+    def execute_partial(x):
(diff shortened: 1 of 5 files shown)


workflow/scripts/parallel_vcf.py
```diff
@@ -0,0 +1,50 @@
...
+
+with tempfile.TemporaryDirectory(dir=tmp_base) as t:
+    def f(x):
+        #print("start", x, file=sys.stderr)
+        cmd = f"bcftools view -r {x} {args.file_vcf} | bcftools view -t {x} | {command} | bcftools view -Ob > {t}/{x}.bcf"
+        os.system(cmd)
+        #print("done", x, file=sys.stderr)
+        return f"{t}/{x}.bcf"
+
+    with Pool(args.threads) as p:
```
johanneskoester, line 40: I don't see a reason to use a threadpool here. Popen is already async, so you can simply call it several times.

workflow/scripts/parallel_vcf.py
```diff
@@ -0,0 +1,50 @@
...
+parser.add_argument("command")
+parser.add_argument("--stepsize", "-s", type=int, default=1000000)
+parser.add_argument("--threads", "-t", type=int, default=8)
+parser.add_argument("--tmp", default=None, help="Define the base directory for the tmp files")
+args = parser.parse_args()
+tmp_base = args.tmp
+stepsize = args.stepsize
+
+# create regions
+f = VariantFile(args.file_vcf)
```
johanneskoester, line 19: Use a with statement here, so that you can avoid the close.

workflow/scripts/parallel_vcf.py
```diff
@@ -0,0 +1,50 @@
...
+regions = []
+for name, contig in f.header.contigs.items():
+    last = 0
+    for x in list(range(0, contig.length, stepsize)):
+        stop = min(last + stepsize - 1, contig.length)
+        regions.append(f"{name}:{last}-{stop}")
+        last = stop+1
```
johanneskoester, line 26: Format with black.

workflow/scripts/parallel_vcf.py
```diff
@@ -0,0 +1,50 @@
...
+
+    with Pool(args.threads) as p:
+        first = True
+        for filename in p.imap(f, regions):
+            print("merge", filename, file=sys.stderr)
+            if first:
+                cmd = f"bcftools view {filename}"
+            else:
+                cmd = f'bcftools view {filename} | grep -v "#"'
+            os.system(cmd)
```
johanneskoester, line 48: Why not use `bcftools concat`?

workflow/scripts/parallel_vcf.py
```diff
@@ -0,0 +1,50 @@
...
+#command = "SnpSift annotate {ss_args} {args.database_vcf} /dev/stdin"
+command = args.command
+
+with tempfile.TemporaryDirectory(dir=tmp_base) as t:
+    def f(x):
+        #print("start", x, file=sys.stderr)
+        cmd = f"bcftools view -r {x} {args.file_vcf} | bcftools view -t {x} | {command} | bcftools view -Ob > {t}/{x}.bcf"
+        os.system(cmd)
+        #print("done", x, file=sys.stderr)
+        return f"{t}/{x}.bcf"
```
johanneskoester, line 38: Use pathlib for joining paths.

workflow/scripts/parallel_vcf.py
```diff
@@ -0,0 +1,50 @@
...
+        stop = min(last + stepsize - 1, contig.length)
+        regions.append(f"{name}:{last}-{stop}")
+        last = stop+1
+f.close() 
+
+#command = "SnpSift annotate {ss_args} {args.database_vcf} /dev/stdin"
+command = args.command
+
+with tempfile.TemporaryDirectory(dir=tmp_base) as t:
+    def f(x):
```
johanneskoester, line 33: Please use readable variable and function names.

workflow/scripts/parallel_vcf.py
```diff
@@ -0,0 +1,50 @@
...
+        #print("start", x, file=sys.stderr)
+        cmd = f"bcftools view -r {x} {args.file_vcf} | bcftools view -t {x} | {command} | bcftools view -Ob > {t}/{x}.bcf"
```
johanneskoester, line 35: * Why two views behind each other?
* Whenever piping between bcftools, one should use -Ou instead of VCF output (uncompressed binary instead of plain text).

johanneskoester:
One other thing. I think it will fail once we have BNDs, won't it? They may consist of multiple records in multiple places. We would need to separate them into another VCF which is processed sequentially, and then merge back afterwards. Just a thing to keep in mind. Can you add a TODO above the rule?

workflow/scripts/parallel_vcf.py
christopher-schroeder: Because you have to call "concat" at the end, when all parts are finished. By doing the code above, you can print (and delete) a part as soon as it is calculated, while everything is still running. This reduces the required temp space.

workflow/scripts/parallel_vcf.py
christopher-schroeder: jub

workflow/scripts/parallel_vcf.py
christopher-schroeder: yes

workflow/scripts/parallel_vcf.py
christopher-schroeder: The reason for the 2 views is rather stupid. When you pass a region r to bcftools and uses the indexed view, then any variant overlapping the start position of r is taken. So even Indels that start outside of r. Of cause, then you have some duplicated variants, if you split up the vcf like I do and there was no flag to prevent this. This is not the case if you use the streaming view with a region (-t). But on the downside the streaming view does not jump right to the region position, but goes through everything before. Therefore the easiest way I could imaging was to first use the indexed view and then eliminating variants, that starts before r by using the streaming view.

christopher-schroeder:
Hm that might be, but I don't even know if the annotation takes the paired BND mates into account. I would guess that SnpSift analyses them separately.

workflow/scripts/parallel_vcf.py
christopher-schroeder: yes

workflow/scripts/parallel_vcf.py
christopher-schroeder: There could be thousands of parts that has to be generated, printed out as soon as they are finished (see comment below for "Why not just use bcftools concat?") and deleted. Everything in order to not screw up the sorting.
How do you manage, that only a limited number of commands are executed at the same time, while everything is print out in order without a threapool and imap? Maybe my skills for parallel execution are not developed enough.

workflow/scripts/parallel_vcf.py
christopher-schroeder: yes

workflow/scripts/parallel_vcf.py
christopher-schroeder: What we could do:
generating a (temporary) named pipe  for each part and at the beginning call bcftools concat on all the pipes. Then filling the pipes on by one and after each fill remove the temporary part. In this way we could make use of concat while still having the same resource- and time-saving saving behavior.

workflow/scripts/parallel_vcf.py
christopher-schroeder: Publishing the script as a wrapper is problematic. Different used commands, in this case "SnpSift annotate" require different environments and with a wrapper you have it fixed. I could create a wrapper for each command, should I?
I think I will create a bioconda package.

johanneskoester:
A note: I am in the process to migrate the pipeline to VEP instead of snpeff /snpsift. VEP has builtin parallelization.
Another reason is that it is in sync with ensembl, always giving you the latest release. And it offers numerous plugins and additional scores. Far beyond nowadays of what snpeff can do.

christopher-schroeder:
Please make sure that VEP also annotated all of our structural variants as they stated
"The Ensembl VEP does not only allow you to annotate short variants, such as SNPs and short insertions or deletions, but also some types of structural variants"

christopher-schroeder:
And most important, can VEP handle Breakend notation?
This seems to be a bit tricky:
https://github.com/Ensembl/ensembl-vep/issues/441

johanneskoester:
Closing in favor of #55.

PR #20, opened 2020-04-07
Limit freebayes by region file

FelixMoelder:
This PR adds new rules for creating a region file from mapped sequence data.
Region files are then forwarded for freebayes variant calling to reduce runtime.

### DIFF
--- a/workflow/rules/candidate_calling.smk
+++ b/workflow/rules/candidate_calling.smk
@@ -13,7 +13,7 @@ rule freebayes:
         extra=config["params"].get("freebayes", ""),
     threads: 100 # use all available cores for calling
     wrapper:
-        "0.50.3/bio/freebayes" #Needs update
+        "0.51.0/bio/freebayes" #Needs update
 
 
 rule delly:
--- a/workflow/rules/candidate_calling.smk
+++ b/workflow/rules/candidate_calling.smk
@@ -2,9 +2,9 @@ rule freebayes:
     input:
         ref="resources/genome.fasta",
         ref_idx="resources/genome.fasta.fai",
+        regions="results/regions/{group}.bed",
         # you can have a list of samples here
         samples=get_group_bams
-        regions="results/regions/{group}.bed"
     output:
         "results/candidate-calls/{group}.freebayes.bcf"
     log:
--- a/workflow/rules/candidate_calling.smk
+++ b/workflow/rules/candidate_calling.smk
@@ -13,7 +13,7 @@ rule freebayes:
         extra=config["params"].get("freebayes", ""),
     threads: 100 # use all available cores for calling
     wrapper:
-        "0.51.2/bio/freebayes" #Needs update
+        "0.51.3/bio/freebayes"
 
 
 rule delly:
--- a/workflow/rules/candidate_calling.smk
+++ b/workflow/rules/candidate_calling.smk
@@ -2,7 +2,7 @@ rule freebayes:
     input:
         ref="resources/genome.fasta",
         ref_idx="resources/genome.fasta.fai",
-        regions="results/regions/{group}.bed",
+        regions=get_regions,
         # you can have a list of samples here
         samples=get_group_bams
     output:
--- a/workflow/rules/candidate_calling.smk
+++ b/workflow/rules/candidate_calling.smk
@@ -13,7 +13,7 @@ rule freebayes:
         extra=config["params"].get("freebayes", ""),
     threads: 100 # use all available cores for calling
     wrapper:
-        "0.51.0/bio/freebayes" #Needs update
+        "0.51.2/bio/freebayes" #Needs update
 
 
 rule delly:
--- a/workflow/rules/candidate_calling.smk
+++ b/workflow/rules/candidate_calling.smk
@@ -4,16 +4,16 @@ rule freebayes:
         ref_idx="resources/genome.fasta.fai",
         # you can have a list of samples here
         samples=get_group_bams
+        regions="results/regions/{group}.bed"
     output:
         "results/candidate-calls/{group}.freebayes.bcf"
     log:
         "logs/freebayes/{group}.log"
     params:
         extra=config["params"].get("freebayes", ""),
-        chunksize=100000
     threads: 100 # use all available cores for calling
     wrapper:
-        "0.50.3/bio/freebayes"
+        "0.50.3/bio/freebayes" #Needs update
 
 
 rule delly:
--- a/workflow/rules/utils.smk
+++ b/workflow/rules/utils.smk
@@ -43,7 +43,7 @@ rule get_covered_regions:
     output:
         temp("results/regions/temp/{sample}.quantized.bed.gz")
     params:
-        prefix={output}.strip(".quantized.bed.gz")
+        prefix=lambda wc, output: output[0].strip(".quantized.bed.gz")
     shadow: "shallow"
     log:
         "logs/bam-regions/{sample}.log"
@@ -54,7 +54,7 @@ rule get_covered_regions:
 
 rule unzip_quantized_regions:
     input:
-        "results/regions/temp/{sample}.quantized.bed.gz
+        "results/regions/temp/{sample}.quantized.bed.gz"
     output:
         "results/regions/{sample}.bed"
     shell:
--- a/workflow/rules/utils.smk
+++ b/workflow/rules/utils.smk
@@ -46,7 +46,7 @@ rule get_covered_regions:
     conda:
         "../envs/bedregions.yaml"
     shell:
-        "samtools view -b {input.bam} | genomeCoverageBed -ibam stdin -g {input.ref} > {output} 2> {log}"
+        "samtools view -b {input.bam} | bedtools bamtobed -i stdin | bedtools merge -i stdin > {output} 2> {log}"
 
 rule merge_covered_regions:
     input:
--- a/workflow/rules/utils.smk
+++ b/workflow/rules/utils.smk
@@ -42,7 +42,7 @@ rule get_covered_regions:
         bam="results/recal/{sample}.sorted.bam",
         bai="results/recal/{sample}.sorted.bam.bai"
     output:
-        temp("results/regions/{sample}.quantized.bed.gz")
+        temp("results/regions/temp/{sample}.quantized.bed.gz")
     params:
         prefix=lambda wc, output: output[0].split(".quantized.bed.gz")[0]
     shadow: "minimal"
--- a/workflow/rules/utils.smk
+++ b/workflow/rules/utils.smk
@@ -44,7 +44,7 @@ rule bam_regions:
     log:
         "logs/bam-regions/{sample}.log"
     conda:
-        "../bedregions.yaml"
+        "../envs/bedregions.yaml"
     shell:
         "samtools view -b {input.bam} | genomeCoverageBed -ibam stdin -g {input.ref} > {output} 2> {log}"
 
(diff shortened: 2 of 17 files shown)


workflow/rules/utils.smk
```diff
@@ -33,4 +33,29 @@ rule tabix_known_variants:
...
+    input:
+        get_group_beds
+    output:
+        "results/regions/{group}.bed"
+    log:
+        "logs/merge-regions/{group}.log"
+    conda:
+        "../envs/bedtools.yaml"
+    shell:
+        "cat {input} | sort -k1,1 -k2,2n | bedtools merge -i stdin > {output} 2> {log} "
```
johanneskoester, line 61: I think the files can be given as direct arguments to sort, no cat needed. Do you have a source explaining why sort is needed here? Can't bedtools do this on its own?

johanneskoester:
Please update to latest master, I just fixed the no space left issue.

workflow/rules/utils.smk
FelixMoelder: Actually the bedtools documentation tells that data needs to be presorted.
The the "Note" at the[documentation](https://bedtools.readthedocs.io/en/latest/content/tools/merge.html)

workflow/rules/utils.smk
```diff
@@ -33,4 +33,29 @@ rule tabix_known_variants:
         get_tabix_params
     cache: True
     wrapper:
-        "0.45.1/bio/tabix"
\ No newline at end of file
+        "0.45.1/bio/tabix"
+
+rule bam_regions:
```
johanneskoester, line 38:
```suggestion
rule get_covered_regions:
```

workflow/rules/utils.smk
```diff
@@ -33,4 +33,29 @@ rule tabix_known_variants:
...
+    output:
+        temp("results/regions/{sample}.bed")
+    log:
+        "logs/bam-regions/{sample}.log"
+    conda:
+        "../envs/bedregions.yaml"
+    shell:
+        "samtools view -b {input.bam} | genomeCoverageBed -ibam stdin -g {input.ref} > {output} 2> {log}"
+
+rule merge_regions:
```
johanneskoester, line 51:
```suggestion
rule merge_covered_regions:
```

workflow/rules/utils.smk
```diff
@@ -33,4 +33,29 @@ rule tabix_known_variants:
...
+        ref="resources/genome.fasta",
+        bam="results/recal/{sample}.sorted.bam"
+    output:
+        temp("results/regions/{sample}.bed")
+    log:
+        "logs/bam-regions/{sample}.log"
+    conda:
+        "../envs/bedregions.yaml"
+    shell:
+        "samtools view -b {input.bam} | genomeCoverageBed -ibam stdin -g {input.ref} > {output} 2> {log}"
```
johanneskoester, line 49: Just to confirm, this will output a bed file with only those regions with at least a coverage of 1?

FelixMoelder: Indeed it won't.
So [bamtobed](https://bedtools.readthedocs.io/en/latest/content/tools/bamtobed.html) might be what we are looking for.

I just tested bamtobed for 10 Reads. The output looks like this:
```
1	21707	21850	M05890:70:000000000-CGM9D:1:1114:20746:10100/1	0	+
1	21707	21850	M05890:70:000000000-CGM9D:1:1114:20746:10100/2	0	-
1	31842	31984	M05890:70:000000000-CGM9D:1:2111:25061:8808/1	0	+
1	31842	31986	M05890:70:000000000-CGM9D:1:2111:25061:8808/2	0	-
1	189965	190072	M05890:70:000000000-CGM9D:1:1105:9369:15622/2	0	+
1	189965	190111	M05890:70:000000000-CGM9D:1:1107:17171:18596/1	23	+
1	189965	190111	M05890:70:000000000-CGM9D:1:1109:2264:16933/2	23	+
1	189965	190072	M05890:70:000000000-CGM9D:1:1105:9369:15622/1	0	-
1	190003	190143	M05890:70:000000000-CGM9D:1:1107:17171:18596/2	23	-
1	190003	190143	M05890:70:000000000-CGM9D:1:1109:2264:16933/1	23	-
```

I guess the first three columns are what we would expect.

FelixMoelder: `samtools view -b <bam> | bedtools bamtobed -i stdin | bedtools merge -i stdin` would end up like this for the example output above: 
```
1	21707	21850
1	31842	31986
1	189965	190143
```

johanneskoester: But bamtobed will generate a huge overhead with one entry per read, as pure text. I think mosdepth is the way to go. It can output regions, that are even quantized, such that we can control the resolution. Probably we can even quantize into `0` and `>0`. https://github.com/brentp/mosdepth

johanneskoester: Wait, maybe bedtools genomecov works as well. https://bedtools.readthedocs.io/en/latest/content/tools/genomecov.html. You can use -max 1, to bin all regions with at least a coverage of 1 into a single bin.

johanneskoester: That would be preferable to mosdepth, cause it is easier to pipe.

FelixMoelder: Using the `-max 1` creates some weird output:

```1	0	248955957	248956422	0.999998
1	1	465	248956422	1.8678e-06
10	0	133797422	133797422	1
11	0	135086622	135086622	1
12	0	133275309	133275309	1
13	0	114364328	114364328	1
14	0	107043718	107043718	1
15	0	101991189	101991189	1
16	0	90338345	90338345	1
17	0	83257441	83257441	1
18	0	80373285	80373285	1
19	0	58617616	58617616	1
2	0	242193529	242193529	1
20	0	64444167	64444167	1
21	0	46709983	46709983	1
22	0	50818468	50818468	1
3	0	198295559	198295559	1
4	0	190214555	190214555	1
5	0	181538259	181538259	1
6	0	170805979	170805979	1
7	0	159345973	159345973	1
8	0	145138636	145138636	1
9	0	138394717	138394717	1
MT	0	16569	16569	1
X	0	156040895	156040895	1
Y	0	57227415	57227415	1
...
```

FelixMoelder: mosdepth requires a bed-file (specified by the `-by` parameter) as input to create the regions?

dlaehnemann: What genomecov does, is give you the number of sites with 0 coverage per chromosome (0 in 2nd column) and as you requested `-max 1`, also the number of sites with coverage 1 or above. Check out the docs for a detailed description of the columns:
https://bedtools.readthedocs.io/en/latest/content/tools/genomecov.html

It seems like your sample only has coverage for very few sites on chromosome 1, which seems weird.

dlaehnemann: If I understand correctly that you want a bed file with all the regions that have read coverage, I think what you want is [`bedtools merge`](https://bedtools.readthedocs.io/en/latest/content/tools/merge.html) that you can also do on BAM files to "merge" all reads into contiguous regions.

dlaehnemann: So basically it should just work on the BAM file without any piping necessary:

```
bedtools merge -i {input.bam} > {output.bed}
```

dlaehnemann: Ah yeah, mosdepth quantization also works! :D

FelixMoelder:
This currently fails as the wrapper turns out to be broken for region files.
https://github.com/snakemake/snakemake-wrappers/pull/86 must first be handled.

FelixMoelder:
@johanneskoester Last commit introduced the shadow directive to remove all output created by mosdepth but the subdirectory still exists after completion. Did I do something wrong there?

workflow/rules/utils.smk
```diff
@@ -33,4 +33,33 @@ rule tabix_known_variants:
...
+        "0.45.1/bio/tabix"
+
+
+rule get_covered_regions:
+    input:
+        ref="resources/genome.fasta",
+        bam="results/recal/{sample}.sorted.bam",
+        bai="results/recal/{sample}.sorted.bam.bai"
+    output:
+        "results/regions/temp/{sample}.quantized.bed.gz"
```
johanneskoester, line 45:
```suggestion
        temp("results/regions/temp/{sample}.quantized.bed.gz")
```

workflow/rules/utils.smk
```diff
@@ -33,4 +33,33 @@ rule tabix_known_variants:
...
+    params:
+        prefix=lambda wc, output: output[0].split(".quantized.bed.gz")[0]
+    shadow: "shallow"
```
johanneskoester, line 48:
```suggestion
    shadow: "minimal"
```

workflow/rules/utils.smk
```diff
@@ -33,4 +33,33 @@ rule tabix_known_variants:
...
+    log:
+        "logs/bam-regions/{sample}.log"
+    conda:
+        "../envs/mosdepth.yaml"
```
johanneskoester, line 52:
```suggestion
        "../envs/mosdepth.yaml"
    group: "covered-regions"
```

workflow/rules/utils.smk
```diff
@@ -33,4 +33,33 @@ rule tabix_known_variants:
...
+        "mosdepth {params.prefix} {input.bam} -q 1: 2> {log}"
+
+
+rule unzip_quantized_regions:
+    input:
+        "results/regions/temp/{sample}.quantized.bed.gz"
+    output:
+        "results/regions/{sample}.bed"
+    log:
+        "logs/unzip_regions/{sample}.log"
```
johanneskoester, line 63:
```suggestion
        "logs/unzip_regions/{sample}.log"
    group: "covered-regions"
```

workflow/rules/candidate_calling.smk
```diff
@@ -2,6 +2,7 @@ rule freebayes:
     input:
         ref="resources/genome.fasta",
         ref_idx="resources/genome.fasta.fai",
+        regions="results/regions/{group}.bed",
```
johanneskoester, line 5: For WGS, the region file is superfluous. Hence, we should make it switchable in the config. If it is switched of, a function here can return `[]`.

FelixMoelder:
I had do implement some changes as freebayes receives a whole group of bam files. Therefore it is necessary to create a bed-file for the whole group.
For now this is a shell-script which can be replaced by an updated snakemake-wrapper.

workflow/rules/utils.smk
```diff
@@ -33,4 +33,37 @@ rule tabix_known_variants:
...
+        bam="results/recal/{sample}.sorted.bam",
+        bai="results/recal/{sample}.sorted.bam.bai"
+    output:
+        temp("results/regions/temp/{sample}.quantized.bed.gz")
+    params:
+        prefix=lambda wc, output: output[0].split(".quantized.bed.gz")[0]
+    shadow: "minimal"
+    log:
+        "logs/bam-regions/{sample}.log"
+    group: "covered-regions"
```
johanneskoester, line 51:
```suggestion
```

Now the group is not a good idea anymore, because all the mosdepth runs of the same group would run in the same job.

workflow/rules/utils.smk
```diff
@@ -33,4 +33,37 @@ rule tabix_known_variants:
...
+
+
+rule merge_regions:
+    input:
+        get_group_beds,
+    output:
+        "results/regions/{group}.bed"
+    log:
+        "logs/unzip_regions/{group}.log"
+    group: "covered-regions"
```
johanneskoester, line 65:
```suggestion
```

PR #24, opened 2020-04-15
Minor refinements

FelixMoelder:
This PR contains two minor changes:

Cutadapt input is now places into a subfolder "sample/unit...". Using `-` as separator may confuse snakemake when deriving the wildcards.

Mark Duplicates is now optional and can now be changed in the config file.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -63,6 +63,9 @@ calling:
           - "PRESENT"
         filter: filtername
 
+duplicates:
+  filtering: activate
+
 annotations:
   dbnsfp:
     activate: false
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -63,7 +63,7 @@ calling:
           - "PRESENT"
         filter: filtername
 
-duplicates:
+remove_duplicates:
   activate: true
 
 annotations:
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -64,7 +64,7 @@ calling:
         filter: filtername
 
 duplicates:
-  filtering: activate
+  activate: true
 
 annotations:
   dbnsfp:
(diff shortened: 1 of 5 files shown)


config/config.yaml
```diff
@@ -63,6 +63,9 @@ calling:
           - "PRESENT"
         filter: filtername
 
+duplicates:
+  activate: true
```
johanneskoester, line 67: Please add this to the schema (workflow/schemas/config.schema.yaml).

workflow/rules/common.smk
```diff
@@ -34,6 +34,20 @@ units = pd.read_csv(config["units"], sep="\t", dtype={"sample_name": str, "unit_
...
+        return "results/dedup/{}.sorted.bam".format(wildcards.sample)
+    else:
+        return "results/mapped/{}.sorted.bam".format(wildcards.sample)
+
+
+def get_recalibrate_quality_bai_input(wildcards):
+    if is_activated("duplicate_filtering"):
+        return "results/dedup/{}.sorted.bam.bai".format(wildcards.sample)
+    else:
+        return "results/mapped/{}.sorted.bam.bai".format(wildcards.sample)
```
johanneskoester, line 48: Use only one function with a keyword argument `bai=False`. This way, it is not so duplicated.

workflow/schemas/config.schema.yaml
```diff
@@ -117,6 +117,12 @@ properties:
       - filter
       - fdr-control
   
+  duplicate_filtering:
```
johanneskoester, line 120:
```suggestion
  remove_duplicates:
```

config/config.yaml
```diff
@@ -63,6 +63,9 @@ calling:
           - "PRESENT"
         filter: filtername
 
+duplicates:
```
johanneskoester, line 66:
```suggestion
remove_duplicates:
```

PR #27, opened 2020-04-27
Replace trimmomatic by ptrimmer

FelixMoelder:
This PR replaces trimmomatic by ptrimmer, as trimming primers using trimmomatic did not work as expected.

Additionally I had to add pipes as pTrimmer asks for input files containing "_R1" and "_R2"-string in the file names.

### DIFF
--- a/workflow/rules/trimming.smk
+++ b/workflow/rules/trimming.smk
@@ -71,7 +71,7 @@ rule pipe_ptrimmer_pe:
     output:
         pipe("results/trimmed/adapters/{sample}/{unit}.paired_R{read}.fastq.gz")
     log:
-        "logs/pipe-fastqs/catadapt/{sample}-{unit}.{read}.{log}"
+        "logs/pipe-fastqs/catadapt/{sample}-{unit}.{read}.log"
     threads: 0 # this does not need CPU
     shell:
         "cat {input} > {output} 2> {log}"
--- a/workflow/rules/trimming.smk
+++ b/workflow/rules/trimming.smk
@@ -14,7 +14,7 @@ rule cutadapt_pipe:
     output:
         pipe('pipe/cutadapt/{sample}/{unit}.{fq}.{ext}')
     log:
-        "logs/pipe-fastqs/{sample}-{unit}.{fq}.{ext}"
+        "logs/pipe-fastqs/catadapt/{sample}-{unit}.{fq}.{ext}"
     wildcard_constraints:
         ext=r"fastq|fastq\.gz"
     threads: 0 # this does not need CPU
@@ -58,6 +58,8 @@ rule pipe_ptrimmer_se:
         "results/trimmed/adapters/{sample}/{unit}.single.fastq.gz"
     output:
         pipe("results/trimmed/adapters/{sample}/{unit}.single_R1.fastq.gz")
+    log:
+        "logs/pipe-fastqs/ptrimmer/{sample}-{unit}.single.log"
     threads: 0 # this does not need CPU
     shell:
         "cat {input} > {output} 2> {log}"
@@ -68,6 +70,8 @@ rule pipe_ptrimmer_pe:
         "results/trimmed/adapters/{sample}/{unit}.{read}.fastq.gz"
     output:
         pipe("results/trimmed/adapters/{sample}/{unit}.paired_R{read}.fastq.gz")
+    log:
+        "logs/pipe-fastqs/catadapt/{sample}-{unit}.{read}.{log}"
     threads: 0 # this does not need CPU
     shell:
         "cat {input} > {output} 2> {log}"
--- a/workflow/rules/trimming.smk
+++ b/workflow/rules/trimming.smk
@@ -80,18 +80,17 @@ rule ptrimmer_se:
     input:
         "results/trimmed/adapters/{sample}/{unit}.single_R1.fastq.gz"
     output:
-        "results/trimmed/primers/{sample}/{unit}.single_trim_R1.fastq.gz",
+        "results/trimmed/primers/{sample}/{unit}.single.fastq.gz",
     log:
         "logs/ptrimmer/{sample}-{unit}.log"
     params:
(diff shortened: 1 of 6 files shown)


workflow/rules/trimming.smk
```diff
@@ -53,39 +53,61 @@ rule cutadapt_se:
...
-        "results/trimmed/primers/{sample}/{unit}.single.fastq.gz",
+        pipe("results/trimmed/adapters/{sample}/{unit}.single_R1.fastq.gz")
+    threads: 0 # this does not need CPU
+    shell:
+        "cat {input} > {output} 2> {log}"
+
+
+rule pipe_ptrimmer_pe:
+    input:
+        "results/trimmed/adapters/{sample}/{unit}.{read}.fastq.gz"
```
johanneskoester, line 70: Let us rather rename the files in the pipeline into R1/R2. This way, we can omit these two rules.

workflow/rules/trimming.smk
```diff
@@ -53,39 +53,61 @@ rule cutadapt_se:
...
+        "logs/ptrimmer/{sample}-{unit}.log"
     params:
-        trimmers=["ILLUMINACLIP:{}:{}:{}:{}".format(*config["primers"]["trimming"]["trimmomatic"].values())],
-        extra="",
-        compression_level="-9"
-    wrapper:
-        "0.51.2/bio/trimmomatic/pe"
+        primers=config["primers"]["trimming"]["primers"],
+        tmp_fq1="results/trimmed/adapters/{sample}/{unit}.paired_trim_R1.fq",
+        tmp_fq2="results/trimmed/adapters/{sample}/{unit}.paired_trim_R2.fq",
```
johanneskoester, line 109: Above, the linter should complain...

johanneskoester: Make sure to remove the uncompressed files again. Once this works, we will create a proper wrapper, that hides such details in a temporary folder.

johanneskoester: Ah, the linter cannot see that, but params.folder can actually a different path than you expect here (due to people using --default-remote-prefix). I suggest to omit above two parameters, and instead use params.folder in the unzip command, plus an infix like `{sample}/{unit}.paired_trim_`, and a suffix R1/R2 directly in the command.

workflow/rules/trimming.smk
FelixMoelder: I tried this at first but I was not able to resolve this as ptrimmer needs the "_R" syntax even for single end mode why is why we also must rename "single" into "_R1".
This way it was not possible to distingish between paired end and single end rules.

workflow/rules/trimming.smk
FelixMoelder: I think

> Make sure to remove the uncompressed files again. Once this works, we will create a proper wrapper, that hides such details in a temporary folder.

The uncompressed files should be deleted by gzip during compression.

FelixMoelder: > Ah, the linter cannot see that, but params.folder can actually a different path than you expect here (due to people using --default-remote-prefix). I suggest to omit above two parameters, and instead use params.folder in the unzip command, plus an infix like `{sample}/{unit}.paired_trim_`, and a suffix R1/R2 directly in the command.

Okay, the three params are removed now.
But I'm not show about how to remove the piping rules. As cutadapt must stick the `single` keyword for single end reads, while ptrimmer asks for `_R1`-syntax for single-end reads.

FelixMoelder:
With the latest changes it was possible to remove at least the piping-rule for pTrimmer paired-end mode.
For single end mode I don't see a nice way as "_R1" needs to be in to file name even for single-end reads which is not possible as cutadapt se and pe would both created the same forward-read output.

workflow/rules/trimming.smk
```diff
@@ -53,49 +53,60 @@ rule cutadapt_se:
...
+    log:
+        "logs/pipe-fastqs/ptrimmer/{sample}-{unit}.single.log"
+    threads: 0 # this does not need CPU
+    shell:
+        "cat {input} > {output} 2> {log}"
+
+
+rule ptrimmer_se:
+    input:
+        "results/trimmed/adapters/{sample}/{unit}.single_R1.fastq.gz"
```
johanneskoester, line 70: Are you sure that it complains about the file name in the single end case?

FelixMoelder: Currently not, but if the input for `ptrimmer_se` becomes `"results/trimmed/adapters/{sample}/{unit}_R1.fastq.gz"` both rules `cutadapt_pe` and `cutadapt_se` need to produce the same output.

FelixMoelder: Changing input to `_R1` for single-end results in the following error:
```
AmbiguousRuleException:
Rules cutadapt_se and cutadapt_pe are ambiguous for the file results/trimmed/adapters/a/lane1_R1.fastq.gz.
Consider starting rule output with a unique prefix, constrain your wildcards, or use the ruleorder directive.
Wildcards:
	cutadapt_se: sample=a,unit=lane1
	cutadapt_pe: sample=a,unit=lane1
Expected input files:
	cutadapt_se: pipe/cutadapt/a/lane1.fq1.fastq pipe/cutadapt/a/lane1.fq2.fastq
	cutadapt_pe: pipe/cutadapt/a/lane1.fq1.fastq pipe/cutadapt/a/lane1.fq2.fastq
Expected output files:
	cutadapt_se: results/trimmed/adapters/a/lane1_R1.fastq.gz results/trimmed/adapters/a/lane1.single.qc.txt
	cutadapt_pe: results/trimmed/adapters/a/lane1_R1.fastq.gz results/trimmed/adapters/a/lane1_R2.fastq.gz results/trimmed/adapters/a/lane1.paired.qc.txt
```

johanneskoester: Add a ruleorder statement: `ruleorder: cutadapt_pe > cutadapt_se`.

PR #29, opened 2020-04-28
Fixing pTrimmer rules

FelixMoelder:
This PR fixes serveral issues causing ptrimmer to fail.

Fixes:

- Wrong output path where gzip can not find input file
- Missing `-s` parameter
- Concatenating of ptrimmer and gzip commands

Improvements:

- Higher compression rate
- Writing stdout and stderr to log-file

Current issues:

- [x] Remove files created by ptrimmer after compression
- [x] `merge_fastqs` seems to be called before both gzip processes have been finished

Currently files created by ptrimmer still exist after compression as writing to stdout keeps the original file. Does anyone know how to remove those without doing so manually?

Should we add two threads as there are two gzip processes?

More about it seems like `merge_fastqs` is called before both gzip processes have been finished as bwa crashes. The issue here is, that one of the merged fastq-mates has an unexpected `end-of-file` error while the the trimmed fastqs are fine. After deleting the merge fastqs the issue is solved as merging runs on the now finished trimmed fastqs.

### DIFF
--- a/workflow/rules/trimming.smk
+++ b/workflow/rules/trimming.smk
@@ -99,8 +99,8 @@ rule ptrimmer_pe:
         "../envs/ptrimmer.yaml"
     shell:
         "ptrimmer -s pair -f {input.r1} -r {input.r2} -a {params.primers} -o {params.output_dir} &> {log} && "
-        "(gzip -c -9 results/trimmed/primers/{wildcards.sample}/{wildcards.unit}_trim_R1.fq > {output.r1} & "
-        "gzip -c -9 results/trimmed/primers/{wildcards.sample}/{wildcards.unit}_trim_R2.fq > {output.r2})"
+        "gzip -c -9 results/trimmed/primers/{wildcards.sample}/{wildcards.unit}_trim_R1.fq > {output.r1} && "
+        "gzip -c -9 results/trimmed/primers/{wildcards.sample}/{wildcards.unit}_trim_R2.fq > {output.r2}"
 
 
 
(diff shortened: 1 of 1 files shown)


johanneskoester:
I guess the last paragraph is solved now? You can gzip inplace and then use mv, or your gzip like now and then use rm afterwards. These are the only two solutions. I would say for now add rm statements.

FelixMoelder:
> I guess the last paragraph is solved now? You can gzip inplace and then use mv, or your gzip like now and then use rm afterwards. These are the only two solutions. I would say for now add rm statements.

Yes, that should be solved now. `rm` has been also added.

workflow/rules/trimming.smk
```diff
@@ -71,33 +71,40 @@ rule ptrimmer_se:
...
+    log:
+        "logs/ptrimmer/{sample}-{unit}.log"
     conda:
         "../envs/ptrimmer.yaml"
     shell:
-        "ptrimmer -f {input} -a {params.primers} -o ./ > {log}"
-        "gzip -c results/trimmed/adapters/{wildcards.sample}/{wildcards.unit}.single_trim_R1.fq > {output}"
+        "ptrimmer -s single -f {input} -a {params.primers} -o {params.output_dir} &> {log} && "
+        "gzip -c results/trimmed/primers/{wildcards.sample}/{wildcards.unit}.single_trim_R1.fq > {output} && "
+        "rm gzip -c results/trimmed/primers/{wildcards.sample}/{wildcards.unit}.single_trim_R1.fq"
```
johanneskoester, line 84: Is this intentional? Looks like a copy-paste error.

FelixMoelder: Indeed. Should be fixed now.

PR #35, opened 2020-05-15
Fix known variants

FelixMoelder:
This PR fixes the current workflow as `get_known_variants` fails due to a missing `build` parameter.

Additionally, all wrappers are updated to `0.55.1`.


FelixMoelder:
We could wait for the [updated samtools index wrapper](https://github.com/snakemake/snakemake-wrappers/pull/103) and merge after updating.

PR #37, opened 2020-05-20
Change targeted regions

FelixMoelder:
This PR changes the way targeted regions are supplied.

Previously for variant calling a bed-file has been created based on the coverage.
Now, an existing bed-file must be defined in the `config.yaml`.

Bugfix:
There also existed a bug causing the workflow to fail when calling based on regions has been deactivated.


johanneskoester:
I have thought about the primer definition. In fact, all we need is a pair of FASTQ files. We can then map them against the used reference genome to generate the intervals and everything needed for ptrimmer. This is much better if the genome changes. And it is also easier to provide, given that there is no standard for primer definition.

FelixMoelder:
It looks like the Linter does not like printing inside shell commands. What is the right way to perform awk printing?

workflow/rules/trimming.smk
```diff
@@ -68,18 +68,18 @@ rule pipe_ptrimmer_se:
...
     params:
-        primers=config["primers"]["trimming"]["primers"],
         output_dir=lambda wc, output: os.path.dirname(output)
     log:
         "logs/ptrimmer/{sample}-{unit}.log"
     conda:
         "../envs/ptrimmer.yaml"
     shell:
-        "ptrimmer -s single -f {input} -a {params.primers} -o {params.output_dir} &> {log} && "
+        "ptrimmer -s single -f {input} -a {input.primers} -o {params.output_dir} &> {log} && "
```
johanneskoester, line 82: `{input}}` will give you the entire input (including primers). Instead, give the fastq input also a name.

workflow/rules/utils.smk
```diff
@@ -36,32 +36,76 @@ rule tabix_known_variants:
...
+    log:
+        "logs/bwa_mem/primers.log"
     params:
-        prefix=lambda wc, output: output[0].split(".quantized.bed.gz")[0]
-    shadow: "minimal"
+        index=lambda w, input: os.path.splitext(input.idx[0])[0],
+        sort="samtools",
+        sort_order="coordinate",
+        extra="-T 10 -k 8 -c 5000"
+
```
johanneskoester, line 52:
```suggestion
```

workflow/rules/utils.smk
```diff
@@ -36,32 +36,76 @@ rule tabix_known_variants:
...
+    input:
+        "resources/genome.fasta.fai"
+    output:
+        "results/regions/genome_regions.bed"
+    params:
+        chroms=config["ref"]["n_chromosomes"]
+    log:
+        "logs/regions/genome_regions.log"
+    script:
+        "../scripts/fasta_generate_genome.py"
```
johanneskoester, line 97: Please choose a more speaking name.

johanneskoester: build-genome-bed? Isn't fai almost a bed?

workflow/scripts/fasta_generate_genome.py
```diff
@@ -0,0 +1,10 @@
+nbr_chromosomes = snakemake.params[0]
+out_file = open(snakemake.output[0], "w")
+with open(snakemake.input[0], "r") as ref_idx:
+    for i in range(nbr_chromosomes):
+        line = ref_idx.readline()
+        fields = line.strip().split("\t")
+        chromosome = fields[0]
+        chromosome_len = fields[1]
+        print("{chr}\t{len}".format(chr=chromosome, len=chromosome_len), file=out_file)
```
johanneskoester, line 9: This is not a bed at all, isn't it?

workflow/rules/utils.smk
```diff
@@ -36,32 +36,76 @@ rule tabix_known_variants:
...
     log:
-        "logs/bam-regions/{sample}.log"
+        "logs/primers/target_regions.log"
     conda:
-        "../envs/mosdepth.yaml"
+        "../envs/bedtools.yaml"
     shell:
-        "mosdepth {params.prefix} {input.bam} -q 1: 2> {log}"
+        "samtools sort -n {input} | bamToBed -i - -bedpe | "
+        "awk '{{print $1 \"\t\" $2 \"\t\" $6}}' | "
```
johanneskoester, line 83: Use cut, looks simpler.

johanneskoester: or even better, xsv

workflow/rules/utils.smk
johanneskoester: maybe this can be a one liner with xsv

config/config.yaml
```diff
@@ -21,9 +21,9 @@ ref:
 primers:
   trimming:
     activate: false
-    # path to amplicon file in the format defined by ptrimmer:
-    # https://github.com/DMU-lilab/pTrimmer
-    primers: "path/to/primer-txt"
+    # path to fasta files containg primer sequences
+    primers_fa1: "path/to/primer-fa1"
+    primers_fa2: "path/to/primer-fa2"
```
johanneskoester, line 26: there should be a separate config in .test that runs the workflow without primer trimming

workflow/scripts/fasta_generate_genome.py
FelixMoelder: You are right. We just need the chromosomes and it lengths here.
So we can just cut the first two lines of the fasta index.

workflow/rules/utils.smk
FelixMoelder: I would prefer using `cut` as this exists by default and we don't need add another dependancy.

FelixMoelder: Additionally, I moved this rule into `build_excluded_regions` as it just reads existing information of the input.

config/config.yaml
FelixMoelder: How does adding an additional config work as the Snakefile links to the existing one?

PR #39, opened 2020-06-08
Filter mapped primers

FelixMoelder:
Currently, the workflow fails in case pairs of primers map to different chromosomes.
To fix this these primers are removed from the final bed file holding the targeted regions.

primers on different chromosomes are written tab-separated into the log holding information about primer name and both chromosomes.

### DIFF
--- a/workflow/scripts/build_target_regions.py
+++ b/workflow/scripts/build_target_regions.py
@@ -1,20 +1,24 @@
+import pandas as pd
+
 log_file = open(snakemake.log[0], "w")
 out = open(snakemake.output[0], "w")
 
-with open(snakemake.input[0], "r") as primer_file:
-    for line in primer_file.readlines():
-        line = line.strip().split("\t")
-        chr_1 = line[0]
-        chr_2 = line[3]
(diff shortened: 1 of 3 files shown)


workflow/scripts/build_target_regions.py
```diff
@@ -0,0 +1,20 @@
+log_file = open(snakemake.log[0], "w")
+out = open(snakemake.output[0], "w")
+
+with open(snakemake.input[0], "r") as primer_file:
+    for line in primer_file.readlines():
```
johanneskoester, line 5:
```suggestion
    for line in primer_file:
```

johanneskoester: Better do this with pandas. Since both input and output are tab separated, should be two or three lines just. The input file could be large, but pandas can easily work on chunks:
https://stackoverflow.com/a/25962187/7070491

FelixMoelder: Done!
While pandas comes along with snakemake should an environment be added to the rule nevertheless?

PR #43, opened 2020-07-14
Replace vep_filter by vembrane

FelixMoelder:
This PR replaces vep_filter by vembrane.

As varlociraptor 1.8 seems to create some invalid bcf records it gets updated to v2.0.
Still, it seems like `varlociraptor preprocess` has a [bug](https://github.com/varlociraptor/varlociraptor/issues/113) causing it to fail for some files.

This has to be fixed before merging.

Todos:

- [x] Test and upgrade vembrane to stable release (at least 0.2)
- [x] Wait for a fixed version of varlociraptor


workflow/envs/varlociraptor.yaml
```diff
@@ -2,4 +2,4 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - varlociraptor =1.8
\ No newline at end of file
+  - varlociraptor =2.0
```
christopher-schroeder, line 5: unfortunately we cannot merge, since varlociraptor 2.0 has bugs, so we have to wait for 2.1

workflow/envs/vembrane.yaml
```diff
@@ -0,0 +1,10 @@
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - pysam =0.16
+  - python =3.7
+  - poetry >=0.12
+  - pyyaml =5.3
+  - pip:
```
christopher-schroeder, line 9: vembrane is available via bioconda now

workflow/rules/filtering.smk
```diff
@@ -8,9 +8,9 @@ rule filter_by_annotation:
     params:
         filter=lambda w: config["calling"]["filter"][w.filter]
     conda:
-        "../envs/vep.yaml"
+        "../envs/vembrane.yaml"
     shell:
-        "(bcftools view {input} | filter_vep --filter \"{params.filter}\" --vcf_info_field ANN --only_matched | bcftools view -Ob > {output}) 2> {log}"
+        "vembrane {input} \"{params.filter}\" --output-fmt bcf --output {output} &> {log}"
```
christopher-schroeder, line 13: the syntax has changed, its now 
vembrane expression input

FelixMoelder: Thanks. Changed that together with preprocessing the input by splitting multi allelic records.

FelixMoelder:
Additional changes: Updated annotating dgiDB using latest rbt version. It is now possible to configure data sources when querying dgiDB.

FelixMoelder:
This should be fine now.
Testing the latest version of varlociraptor on real data did not showed any issues anymore while processing delly files.

christopher-schroeder:
unfortunately it shows with gridss and unpaiired breakends

FelixMoelder:
> unfortunately it shows with gridss and unpaired breakends

But does this effect the current pipeline? As far as I know we currently do not support gridss and breakends.
Therefore, I don't see any reason waiting for another varlociraptor release.

christopher-schroeder:
No you are right, this does not effect the current pipeline. But afaik the only reason for varlociraptor 2.0 is breakend support. So we want it to be able to replace delly by gridss.

workflow/envs/varlociraptor.yaml
christopher-schroeder: varlociraptor 2.0.0 is buggy, we have to set this to >=2.0.1 <2.1 or =2.0.1

config/config.yaml
```diff
@@ -67,7 +67,7 @@ calling:
     # and can be referred in FDR control below to generate calls for different events.
     # In particular, you can also filter by ID or dbsnp annotations here.
     # See http://snpeff.sourceforge.net/SnpSift.html#filter
-    filtername: "ANN[*].IMPACT has 'MODERATE'"
+    filtername: \"ANN['IMPACT'] == 'MODERATE'\"
```
johanneskoester, line 70: Please move the quotation into the rule (snakemake has `{params.filter:q}` which will take care of quoting and escapes. No need to bug people in here with escaping quotes.

workflow/rules/filtering.smk
```diff
@@ -1,16 +1,29 @@
...
     log:
         "logs/filter-calls/annotation/{group}.{filter}.log"
     params:
         filter=lambda w: config["calling"]["filter"][w.filter]
     conda:
-        "../envs/vep.yaml"
+        "../envs/vembrane.yaml"
     shell:
-        "(bcftools view {input} | filter_vep --filter \"{params.filter}\" --vcf_info_field ANN --only_matched | bcftools view -Ob > {output}) 2> {log}"
+        "vembrane \"{params.filter}\" {input} --output-fmt bcf --output {output} &> {log}"
```
johanneskoester, line 26:
```suggestion
        "vembrane {params.filter:q} {input} --output-fmt bcf --output {output} &> {log}"
```

PR #54, opened 2020-10-09
Replace igv-report by rbt vcf-report

FelixMoelder:
As the title says the igv-report gets replaced by vcf-report.
Additionally, oncoprint have been removed as it is already included in vcf-report.


workflow/rules/report.smk
```diff
@@ -1,21 +1,19 @@
...
-    log:
-        "logs/igv-report/{group}.{event}.log"
+        report(directory("results/vcf-report/{batch}.{event}/"), htmlindex="index.html", caption="../report/calls.rst", category="Variant calls")
     params:
-        os.path.join(os.path.dirname(workflow.snakefile), "resources/igv-report-template.html")
+        bcfs=lambda w: expand("{group}=results/merged-calls/{group}.{event}.fdr-controlled.bcf", group=get_report_batch(w), event=w.event),
+        bams=lambda w: get_batch_bams(w, True),
+        format_field = "-f DP AF OBS",
+        info_field = "-i PROB_FFPE_ARTIFACT PROB_PRESENT PROB_ARTIFACT PROB_ABSENT",
+        template = "-j workflow/resources/custom-table-report.js"
```
johanneskoester, line 13: You need to compose the file path relative to __file__, otherwise it is interpreted as relative to the working directory, which can be somewhere else than the sources (e.g. by using the --directory flag).

johanneskoester (changes requested):

workflow/rules/report.smk
```diff
@@ -1,21 +1,19 @@
...
-        report("results/igv-report/{group}.{event}.html", caption="../report/calls.rst", category="Variant calls")
-    log:
-        "logs/igv-report/{group}.{event}.log"
+        report(directory("results/vcf-report/{batch}.{event}/"), htmlindex="index.html", caption="../report/calls.rst", category="Variant calls")
     params:
-        os.path.join(os.path.dirname(workflow.snakefile), "resources/igv-report-template.html")
+        bcfs=lambda w: expand("{group}=results/merged-calls/{group}.{event}.fdr-controlled.bcf", group=get_report_batch(w), event=w.event),
+        bams=lambda w: get_batch_bams(w, True),
+        format_field = "-f DP AF OBS",
+        info_field = "-i PROB_FFPE_ARTIFACT PROB_PRESENT PROB_ARTIFACT PROB_ABSENT",
```
dlaehnemann, line 12: As `event` is a wildcard in this rule, I would use it in specifying which INFO fields to use, here. Users might change the name of the event in the `config.yaml`:
```suggestion
       info_field = lambda w:
           "-i PROB_FFPE_ARTIFACT PROB_{event} PROB_ARTIFACT PROB_ABSENT".format(
               event=w.event.upper()
           ),
```

FelixMoelder: Thanks! That's a good point. Only the expression needs to be a bit diffrent as the varlociraptor events need to be extracted from the config.
It should be something like `lambda wc: "PROB_FFPE_ARTIFACT {} PROB_ARTIFACT PROB_ABSENT".format(' '.join(expand("PROB_{events}", events=config["calling"]["fdr-control"]["events"][wc.event]["varlociraptor"])))`

dlaehnemann: Ah, yes. I was still too focused on my use case. It's even better this way, more general!

workflow/rules/report.smk
```diff
@@ -1,21 +1,20 @@
-rule igv_report:
+rule vcf_report:
     input:
-        bcf="results/merged-calls/{group}.{event}.fdr-controlled.bcf",
         ref="resources/genome.fasta",
-        ref_idx="resources/genome.fasta.fai",
-        bams=lambda w: get_group_bams(w),
-        bais=lambda w: get_group_bams(w, bai=True)
+        bams=lambda w: get_batch_bams(w),
```
johanneskoester, line 4:
```suggestion
        bams=get_batch_bams,
```

workflow/rules/report.smk
```diff
@@ -1,21 +1,20 @@
...
-        report("results/igv-report/{group}.{event}.html", caption="../report/calls.rst", category="Variant calls")
-    log:
-        "logs/igv-report/{group}.{event}.log"
+        report(directory("results/vcf-report/{batch}.{event}/"), htmlindex="index.html", caption="../report/calls.rst", category="Variant calls")
     params:
-        os.path.join(os.path.dirname(workflow.snakefile), "resources/igv-report-template.html")
+        bcfs=lambda w: expand("{group}=results/merged-calls/{group}.{event}.fdr-controlled.bcf", group=get_report_batch(w), event=w.event),
+        bams=lambda w: get_batch_bams(w, True),
+        format_field = "DP AF OBS",
+        template = Path(Path(workflow.snakefile).parent, "resources/custom-table-report.js"),
```
johanneskoester, line 12:
```suggestion
        template = Path(workflow.snakefile).parent / "resources/custom-table-report.js",
```

workflow/rules/report.smk
```diff
@@ -1,21 +1,20 @@
...
+        max_read_depth = config["report"]["max_read_depth"],
+        js_files = Path(Path(workflow.snakefile).parent, "resources/math.min.js")
```
johanneskoester, line 14:
```suggestion
        js_files = Path(workflow.snakefile).parent / "resources/math.min.js"
```

workflow/rules/report.smk
```diff
@@ -1,21 +1,20 @@
...
     conda:
-        "../envs/igv-reports.yaml"
+        "../envs/rbt.yaml"
     shell:
-        "bcftools view {input.bcf} > {output}.vcf;"
-        "create_report {output}.vcf {input.ref} --flanking 100 "
-        "--info-columns ANN dgiDB_drugs cosmic_LEGACY_ID --info-columns-prefixes PROB_ --sample-columns DP AF OBS"
-        " --template {params} --tracks {input.bams} --output {output} --standalone 2>&1 > {log}; "
-        "rm {output}.vcf"
+        "rbt vcf-report {input.ref} --bams {params.bams} --vcfs {params.bcfs} -f {params.format_field} -i PROB_* -j {params.template} -d {params.max_read_depth} -l {params.js_files} -- {output}"
```
johanneskoester, line 20:
```suggestion
        "rbt vcf-report {input.ref} --bams {params.bams} --vcfs {params.bcfs} -f {params.format_field} "
        "-i PROB_* -j {params.template} -d {params.max_read_depth} -l {params.js_files} -- {output}"
```

johanneskoester: Further, please always use the long option names instead of the short ones (readability).

johanneskoester (changes requested):
Awesome! A few comments below.

workflow/rules/report.smk
johanneskoester: Looks to me as this suggestion hasn't been applied.

johanneskoester: The one with the two-liner.

johanneskoester: Wait I'll do it.

johanneskoester (approved):

PR #57, opened 2020-11-13
format workflow with snakefmt

johanneskoester:


dlaehnemann:
I think this is long outdated, with formatting checks even included in the CI nowadays...

PR #61, opened 2020-11-26
Indel realignment, samtools calmd and mapdamage2 recalibration

fgvieira:
Dear all,

I am not sure if this pipeline is intended for shared development and use, or more like an example pipeline for people to get inspiration from but, in case of the former, I've implemented `GATK indel realignment`, `samtools calmd` and `mapdamage2` recalibration.

I've also made a couple more minor tweaks:

- Added `workdir` and fix to avoid conda R conflicts
- Fixed issue with optional parameters (when they are not present in `config.yaml`)
- Switched `picard` `MarkDuplicates` to the newer `MarkDuplicatesWithMateCigar`
- Added PE SRA samples to tests (the ones before were SE)
- Removed unused conda `yaml` environment files, and switched `vembrane` rules to use the wrappers


fgvieira:
There is still a linting error, but not sure on how best to fix it. A while ago I came across this problem and mentioned it in another [issue](https://github.com/snakemake/snakemake-wrappers/pull/257#issuecomment-728901521).

Because of possible `R` conflicts when using `conda`, the fix was to check if `R_LIBS` exists and remove it. However, `snakemake` wants to check if all env variables exist before using them; which makes sense but just not in this specific case. :smile:

johanneskoester:
Hi @fgvieira, and thanks a lot!
I have some questions:
1. what is the rationale for indel realignment? Varlociraptor realigns reads against indels internally, so I don't think it is necessary as an additional step.
2. very interesting to see calmd and mapdamage2 based base recalibration. Can you comment on their effects compared to GATK BQSR?

fgvieira:
As a disclaimer, I must say that these change were done with the type of data that I work with in mind (ancient, degraded, low coverage, non-model organisms) but they can all be enabled/disabled from the config file.

1. When the user only wants the CRAM/BAM file to be used on other programs/pipelines; that was also why I made it optional and controlled by the config. I know it is not supported now, but that would actually be one of my next suggestions: allow the user to choose whether to generate only CRAMs, `raw` VCFs (before varlociraptor), `annotated` VCFs (before varlociraptor), or combined VCFs (with varlociraptor).

2. The problem with `GATK BQSR` is that I do not have a set of known variants (non-model organisms). On top of that, with aDNA you usually only have one sample/library/lane, so also not many features.


PS - what about the `R_LIBS` issue, do you have any idea how we could fix it?

johanneskoester:
It might make sense to bake the R_LIBS handling into Snakemake itself. However, I would not want to completely remove it because it is a useful tool for people who know what they are doing. Maybe it could be removed by default when having a conda env, but that behavior could be deactivated again by setting a flag.

johanneskoester:
Regarding 1.:
Understood, but I'd like to keep this out of this pipeline. Instead, I'd rather propose to change the input definition such that it becomes possible to alternatively specify a bam in units.tsv, which would then bypass the preprocessing and directly start with the calling.

Regarding 2.:
Very good point, makes a lot of sense to have it then.

johanneskoester:
> It might make sense to bake the R_LIBS handling into Snakemake itself. However, I would not want to completely remove it because it is a useful tool for people who know what they are doing. Maybe it could be removed by default when having a conda env, but that behavior could be deactivated again by setting a flag.

Done, see here: https://github.com/snakemake/snakemake/pull/812.

fgvieira:
1. I agree that it would also be nice to allow for BAM input, but that would not solve the cases when the user only wants a BAM file to use with other programs/pipelines. I also don't think that many users will already have the BAM file... most of the time the user will want to start from the FASTQ files and get either `BAM`, `BAM+VCF`, `BAM+VCF+annotVCF`, ... no?
2. Great!

And great to see that the `R_LIBS` issue has been fixed! :smile:

jafors:
>     1. I agree that it would also be nice to allow for BAM input, but that would not solve the cases when the user only wants a BAM file to use with other programs/pipelines. I also don't think that many users will already have the BAM file... most of the time the user will want to start from the FASTQ files and get either `BAM`, `BAM+VCF`, `BAM+VCF+annotVCF`, ... no?

I'm just sliding in your discussion, if that's ok. ;)
I agree that some use-cases might just want a BAM or an unannotated VCF, I already used these "configurable steps" in one of my pipelines myself. 
I also agree that it would not make much sense to have separate pipelines for "Mapping", "Mapping+(Candidate)Calling", "Mapping+Calling+Annotation", etc.
On the other hand, you maybe don't want to stick you workflow full of "exit doors" after every step. But I feel this is a topic to be discussed, since it would enhance modularity of workflows without rebuilding every step over and over. Maybe the "meta-wrapper" or sub-workflow concept could be enhanced in this direction, to easily allow "clicking together" a pipeline with exactly the steps in need.

fgvieira:
I agree that it does not make sense to have "exit doors" everywhere, but at least on the main steps (e.g.):
- BAM
- raw VCF
- annot VCF
- varlociraptor combined VCF

this, combined with config options to opt-out of specific steps (e.g. gatk bqsr) would make the pipeline quite flexible.

PR #66, opened 2021-03-05
Fix input of varlociraptor preprocess for Tumor/Normal Samples

FelixMoelder:
Currently the pipeline fails in case there are normal samples in addition to tumor samples.
This is due to a wrong input getter function and should be fixed now.


FelixMoelder:
@johanneskoester bump ;)

johanneskoester (approved):

PR #67, opened 2021-04-01
Bump snakemake min_version to 6.0.5

tedil:
Since introduction of the scattergather parallelization in this workflow, snakemake was unable to build the DAG due to a stackoverflow occurring because of a repeating wildcard, or more specifically: `{caller}` ended up as part of `{scatteritem}`, leading to the prefix "freebayes." being repeated over and over again. This PR should fix that.
This PR also unifies the use of `{ext}`, which now always has a leading dot in the patterns, i.e. `.{ext}`.
(I also had repeating `.sorted` suffixes, not sure which of these two changes fixed that.)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Lint workflow
-      uses: snakemake/snakemake-github-action@v1.16.0
+      uses: snakemake/snakemake-github-action@v1.17.0
       with:
         directory: .
         snakefile: workflow/Snakefile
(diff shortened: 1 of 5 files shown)


tedil:
Or perhaps `{caller}` should be part of `{group`}? I'm not quite sure on the semantics here.

dlaehnemann:
> Or is caller supposed to be part of scatteritem at that point?
> Or maybe some wildcard constraints are insufficient/incorrect?

If I understand correctly, this rule will concat the results of the different used callers into one file, so the output should not contain the caller. Thus, I think wildcard constraints on `{group}` (which already exists) and `{scatteritem}` should do the job. For the latter, [according to the scatter-gather docs of snakemake](https://snakemake.readthedocs.io/en/latest/snakefiles/rules.html#defining-scatter-gather-processes), this should do the trick:
```
\d+-of-\d+
```

tedil:
> > Or is caller supposed to be part of scatteritem at that point?
> > Or maybe some wildcard constraints are insufficient/incorrect?
> 
> If I understand correctly, this rule will concat the results of the different used callers into one file, so the output should not contain the caller. 

Ah, yes, that makes sense. I reverted the respective change.

> Thus, I think wildcard constraints on `{group}` (which already exists) and `{scatteritem}` should do the job. For the latter, [according to the scatter-gather docs of snakemake](https://snakemake.readthedocs.io/en/latest/snakefiles/rules.html#defining-scatter-gather-processes), this should do the trick:
> 
> ```
> \d+-of-\d+
> ```

rules/common.smk actually has a comment about that near the part where the wildcard coinstraints are defined: `#Todo replace regex by "\d+-of-\d`plus`" when linter allows it`.
So, with
```
wildcard_constraints:
    […]
    scatteritem="\d+-of-\d+",
```
the repetition issues seem to be solved.

tedil:
Okay, so according to @johanneskoester the scattergather constraint now is enforced by snakemake itself, and one should require min_version(6.0.5) in the pipeline instead, which is now basically the only thing this PR does (plus the dot before ext thing).

dlaehnemann (approved):
LGTM (and it's @johanneskoester's suggestion, after all;)

johanneskoester (approved):

PR #78, opened 2021-06-07
default to autodetect primer library size (and error) with yara + value docs

dlaehnemann:
Set the primer trimming default configuration mode to autodetect primer library size (and error) with yara. Document that this is done by setting the `config.yaml` values to 0.


FelixMoelder (approved):
LGTM. Thanks!

PR #80, opened 2021-06-17
Handle single primer-trimming

FelixMoelder:
These are first adjustment of the pipeline when receiving just one primer for trimming.
In this case primers can not be used for region targeting and a separat bed-file is required.
As primers/roi-files often hold coordinates of hg19 a liftover chain url can be defined allowing to perform a liftover for targeted regions.

This will be finalized as soon as fgbio is capable of handling single primers.


workflow/rules/common.smk
```diff
@@ -231,6 +231,22 @@ def get_varlociraptor_preprocessing_input(wildcards, bai=False):
...
+def get_primer_bed():
+    if config["primers"]["trimming"].get("primers_fa2", ""):
+        return "results/primers/primers.bedpe"
+    else:
+        return "results/primers/primers.bed"
+
+
+def get_target_regions():
+    if config["target_regions"].get("bed", ""):
+        if ["target_regions"].get("liftover_chain", ""):
```
dlaehnemann, line 243: If possible, I'd specify the genome version of the `"target_regions"` file here and then automatically download the correct liftover file based on this genome versions and the genome version used for mapping in the workflow.

dlaehnemann (changes requested):
This generally looks good, only one bit where I would prefer a bit more of an automagic behaviour... :mage:

workflow/rules/common.smk
FelixMoelder: Should be possible as long the Syntax would be like `<source>To<target>` as that seems to the the syntax which is used by [ucsc](https://hgdownload.soe.ucsc.edu/goldenPath/hg19/liftOver/)

dlaehnemann: Ah right, they have their own genome version naming scheme. So this will not go together with our genome version specification. So either we map genome version specifications from one to the other, but would also have to do this for other species, or we do one of these:

1. The user does have to specify the liftover file, like you suggested.
2. We find another smart way of determining the regions of interest automatically. Maybe something that uses `mosdepth` to get the regions of the genome that are covered, and of those select only the regions that overlap with at least one of the primers?

dlaehnemann: Yeah, it's a good start. But the danger is that it goes out of sync with the general genome specified for the workflow. Maybe something to discuss with Johannes at some point. But we can leave it like this, for now, and touch it again once we have fgbio working for single primers.

FelixMoelder:
This won't run until single primer handling is included in fgbio (see https://github.com/fulcrumgenomics/fgbio/pull/681).

FelixMoelder:
The last commit changed the order of some rules in the workflow.
Previously, it was not possible to enable consensus read calculation and primer trimming as only capable of trimming paired end data which does not exist after calculation of consensus reads.
This is now solved by performing primer trimming first.
Moreover the workflow preferred consensus calculation over duplicate marking/deduplication.
As marking duplicates can be applied as a preprocessing step for merging duplicates into a single read both rules will now be applied if enabled.

Base recalibration will now be applied after trimming, deduplication and consensus calculation.

workflow/envs/filter_reads.yaml
```diff
@@ -0,0 +1,5 @@
+channels:
+  - conda-forge
+dependencies:
+  - rust-script>=0.17.0
+  - rust
```
mbargull, line 5: maybe try
```suggestion
  - rust
  - compiler-c
```
here to get a `${CC}` which `bzip2-sys`'s build can use.

workflow/scripts/filter_primers.rs
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env rust-script
```
tedil, line 1:
```suggestion
```
you don't actually need the shebang when using rust-scripts in snakemake contexts.

FelixMoelder:
This PR is ready now but we still need to wait for a [new release](https://github.com/fulcrumgenomics/fgbio/issues/715) of fgbio before merging.

workflow/envs/filter_reads.yaml
```diff
@@ -0,0 +1,9 @@
...
+  - c-compiler
+  - cmake
+  - make
+  - perl
```
johanneskoester, line 9: Please also pin the other versions here.

workflow/envs/bowtie.yaml
```diff
@@ -0,0 +1,6 @@
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - bowtie
```
johanneskoester, line 5: Please pin bowtie.

workflow/rules/calling.smk
```diff
@@ -37,8 +37,8 @@ rule varlociraptor_preprocess:
         ref="resources/genome.fasta",
         ref_idx="resources/genome.fasta.fai",
         candidates=get_candidate_calls(),
-        bam=lambda w: get_sample_bam(w),
-        bai=lambda w: get_sample_bam(w, bai=True),
+        bam="results/recal/{sample}.sorted.bam",
+        bai="results/recal/{sample}.sorted.bai",
```
johanneskoester, line 41: No need for the function anymore?

johanneskoester (changes requested):
Epic work! Please find my comments below.

workflow/rules/calling.smk
FelixMoelder: Actually no. This had to be removed due to restructuring.
Previously the input bams for `varlociraptor_preprocess` were either the recalibrated bams or alternatively output of primer trimming. So, recalibration happened before trimming.
As there were some incompatibilities between newly introduced steps base recalibration has moved to the end of the preprocessing workflow so the input needed by `varlociraptor_preprocess` will always be the recalibrated one.

workflow/Snakefile
```diff
@@ -45,6 +45,7 @@ if is_activated("report/stratify"):
 rule all:
     input:
         get_final_output(),
+        expand("results/final-calls/{group}.annotated.bcf", group=groups),
```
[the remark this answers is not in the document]
FelixMoelder: Nope, this was just left behind from testing.

config/config.yaml
```diff
@@ -22,6 +22,9 @@ primers:
     # path to fasta files containg primer sequences
     primers_fa1: "path/to/primer-fa1"
     primers_fa2: "path/to/primer-fa2"
+    # optional primer file allowing to define primers per sample
+    # overwrites primers_fa1 and primers_fa2
```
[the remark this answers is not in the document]
johanneskoester: Ah, and then in the sample sheet you define the panel? Is this already documented?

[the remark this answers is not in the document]
johanneskoester: THis should be explained in depth in config/README.md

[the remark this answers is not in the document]
FelixMoelder: Good point. This appeared intuitive to me but it is not.

config/README.md
```diff
@@ -27,3 +27,12 @@ Missing values can be specified by empty columns or by writing `NA`. Lines can b
 Varlociraptor supports integrated uncertainty aware calling and filtering of variants for arbitrary scenarios. These are defined as so-called scenarios, via a [variant calling grammar](https://varlociraptor.github.io/docs/calling#generic-variant-calling).
 * For each group, a scenario is rendered via [Jinja](https://jinja.palletsprojects.com).
 * Therefore, edit the template scenario (`scenario.yaml`) according to your needs. The sample sheet is available for jinja rendering as a pandas data frame in the variable `samples`. This allows to customize the scenario according to the contents of the sample sheet. You can therefore add additional columns to the sample sheet (e.g. purity) and access them in the scenario template, in order to pass the information to Varlociraptor.
+
+# Primer trimming
+
+When primer trimming is enable primers have to be defined either directly in the `config.yaml` or in a seperate tsv-file.
```
johanneskoester, line 33:
```suggestion
When primer trimming is enabled, primers have to be defined either directly in the `config.yaml` or in a seperate tsv-file.
```

config/README.md
```diff
@@ -27,3 +27,12 @@ Missing values can be specified by empty columns or by writing `NA`. Lines can b
...
+Defining primers directly in the config file is prefered when all samples come from the same primer set.
+In case of different panels, primers have to be set panel-wise in a seperate tsv-file.
+For each panel the following columns need to be set: `panel`, `fa1` and `fa2` (optional).
+Additionally, for each sample the corresponding panel must be defined in `samples.tsv` (column `panel`).
+For single primer trimming only the first entry in the config (respective in the tsv file) needs to be defined.
```
johanneskoester, line 38:
```suggestion
For single primer trimming only, the first entry in the config (respective in the tsv file) needs to be defined.
```

config/README.md
```diff
@@ -27,3 +27,12 @@ Missing values can be specified by empty columns or by writing `NA`. Lines can b
 Varlociraptor supports integrated uncertainty aware calling and filtering of variants for arbitrary scenarios. These are defined as so-called scenarios, via a [variant calling grammar](https://varlociraptor.github.io/docs/calling#generic-variant-calling).
 * For each group, a scenario is rendered via [Jinja](https://jinja.palletsprojects.com).
 * Therefore, edit the template scenario (`scenario.yaml`) according to your needs. The sample sheet is available for jinja rendering as a pandas data frame in the variable `samples`. This allows to customize the scenario according to the contents of the sample sheet. You can therefore add additional columns to the sample sheet (e.g. purity) and access them in the scenario template, in order to pass the information to Varlociraptor.
+
+# Primer trimming
```
johanneskoester, line 31: Add an introductory sentence mentioning that the pipeline allows to trim amplicon primers (both ends of a fragment and only one single end of a fragment). Specify whether the latter is supposed to be the left or the right end.

johanneskoester (changes requested):

config/README.md
```diff
@@ -30,9 +30,11 @@ Varlociraptor supports integrated uncertainty aware calling and filtering of var
 
 # Primer trimming
 
-When primer trimming is enable primers have to be defined either directly in the `config.yaml` or in a seperate tsv-file.
+For panel data the pipeline allows trimming of amplicon primers on both ends of a fragment but also on a single end only. 
+In case of single end primers these are suppost to be located at the left end of a read.
```
johanneskoester, line 34: supposed

config/README.md
```diff
@@ -30,9 +30,11 @@ Varlociraptor supports integrated uncertainty aware calling and filtering of var
...
+When primer trimming is enabled primers have to be defined either directly in the `config.yaml` or in a seperate tsv-file.
```
johanneskoester, line 35: enabled,

johanneskoester (changes requested):

johanneskoester (approved):

PR #85, opened 2021-07-14
Limit freebayes/delly to regions

FelixMoelder:
This PR adds rules for limiting freebayes and delly to covered regions, only.


workflow/rules/regions.smk
```diff
@@ -0,0 +1,47 @@
+rule build_sample_regions:
+    input:
+        get_varlociraptor_preprocessing_input,
```
FelixMoelder, line 3: As the name says this was previously only used for getting varlociraptor input.
To prevent confusion this function should probably get a more general name.

johanneskoester: Fixed with commit 93bd785.

PR #98, opened 2021-11-22
Annotate umis

FelixMoelder:
Pipeline expansion for annotation of molecular barcodes (UMIs).
UMIs will be annotated right after read mapping.
In case of annotation a the `BARCODE`-parameter will be set when executing `picardtools MarkDuplicates`.
This allows building consensus reads by UMI as duplicates are now grouped by UMI.


workflow/rules/plugins.smk
```diff
@@ -0,0 +1,29 @@
+rule download_revel:
+    output:
+        temp("resources/revel_scores.zip"),
+    log:
+        "logs/vep_plugins/download_revel.log",
+    shell:
+        "curl https://rothsj06.u.hpc.mssm.edu/revel-v1.3_all_chromosomes.zip -o {output} &> {log}"
```
FelixMoelder, line 7: This url comes from a dynamic google link. Let's how this never changes...

workflow/rules/ref.smk
```diff
@@ -101,6 +101,7 @@ rule get_vep_cache:
 rule get_vep_plugins:
     output:
         directory("resources/vep/plugins"),
+        "resources/vep/plugins/LoFtool_scores.txt",
```
FelixMoelder, line 104: This is always part of the VEP plugins. But we need to add this to the input of `annotate_variants` for correct handling of the `LoFtool`-plugin.

workflow/rules/annotation.smk
```diff
@@ -22,6 +22,9 @@ rule annotate_variants:
         calls="results/calls/{group}.{scatteritem}.bcf",
         cache="resources/vep/cache",
         plugins="resources/vep/plugins",
+        REVEL=lambda wc: get_plugin_aux("REVEL"),
+        revel_tbi=lambda wc: get_plugin_aux("REVEL", True),
+        LoFtool=lambda wc: get_plugin_aux("Loftool"),
```
FelixMoelder, line 27: These are case sensitive as the VEP snakemake-wrapper is looking for plugin aux-files in the input by looking for the plugin-name.
Alternatively, we could set this to lowercase in the wrapper but this might lead to confusion for users.
Any suggestions?

FelixMoelder:
This PR has already been tested on real data.
Before merging we should probably wait for a new varlociraptor release.

config/config.yaml
```diff
@@ -114,11 +114,12 @@ calling:
...
         filter: myfilter
 
-# Will be ignored if calc_consensus_reads is activated
+# If calc_consensus_reads is activated duplicates will be merged
 remove_duplicates:
   activate: true
 
-# Experimental: calculate consensus reads from PCR duplicates
+# Experimental: calculate consensus reads from PCR duplicates.
+# if 'remove_duplicates' is deactivate only overlapping pairs will be merged
```
johanneskoester, line 122: Please also add the revel setting to this template config (commented out).

johanneskoester (changes requested):
Nice and clean! Just one comment below.

config/config.yaml
FelixMoelder: I guess you mean just adding `REVEL` as a plugin in the config?
This is done now.

workflow/rules/common.smk
```diff
@@ -486,26 +484,14 @@ def get_report_batch(wildcards):
...
             vartype=["SNV", "INS", "DEL", "MNV", "BND", "INV", "DUP", "REP"],
-            filter=config["calling"]["fdr-control"]["events"][wildcards.event][
-                "filter"
-            ],
         )
 
     return inner
 
 
-def get_merge_calls_input_report(wildcards, ext="bcf"):
```
FelixMoelder, line 500: I removed this as this seems to be dead code. At least I could not find where this function is called.

.test/config-sra/config.yaml
```diff
@@ -48,7 +48,8 @@ calling:
       present:
         varlociraptor: 
           - "PRESENT"
-        filter: moderate
+        filter:
+          - moderate
```
johanneskoester, line 52: For convenience, it should still be possible to put just a string here without a list. Can you keep it that way for one of the tests?

workflow/rules/common.smk
```diff
@@ -544,11 +540,17 @@ def get_filter_targets(wildcards, input):
         return ""
 
 
+def get_annotation_filter(wildcards):
+    filter = config["calling"]["fdr-control"]["events"][wildcards.event]["filter"]
```
johanneskoester, line 544: Allow filter to be just a string here (for convenience and compatibility).

workflow/schemas/units.schema.yaml
```diff
@@ -23,6 +23,9 @@ properties:
   adapters:
     type: string
     description: cutadapt adapter trimming settings to use (see https://cutadapt.readthedocs.io)
+  umis:
+    type: string
+    description: path to FASTQ file containing read names and corresponding molecular barcodes (leave emtpy if not exising)
```
johanneskoester, line 28: Shouldn't there also be an update to config.schema.yaml, for the filter changes?

johanneskoester (changes requested):

.test/config-sra/config.yaml
FelixMoelder: Good point. I just reversed one test to cover this.

workflow/rules/common.smk
FelixMoelder: This should work now.

johanneskoester (approved):

PR #105, opened 2022-01-08
fix: trim blocks in scenario rendering

johanneskoester:


johanneskoester:
Closing in favor of YTE based rendering as it is already implemented.

PR #106, opened 2022-01-20
Fix vcf-report

FelixMoelder:
It looks like the annotation field created by VEP has received a new value.
Therefore plots are failing as the index needs to be adjusted.

To prevent wrong indexing in the future it would probably help if `rbt vcf-report` saves the values of the ANN-field in a dictionary for directly accessing the correct values by name.


johanneskoester (approved):

johanneskoester:
Thanks! Indeed we should find a solution for that sooner or later. But the variant reporting is kind of in a flow currently anyway. I think about moving it to datavzrd once that is ready (will be more flexible and easier to adapt). Let's see.

PR #111, opened 2022-02-16
improvement: refactored primer filtering

FelixMoelder:
As the rust script for primer filtering depends htslib it may happen that compiling the rust script fails.
To circumvent this `rust-htslib` has been replaced by `noodles`.

### DIFF
--- a/workflow/scripts/filter_primers.rs
+++ b/workflow/scripts/filter_primers.rs
@@ -4,66 +4,105 @@
 //!
 //! ```cargo
 //! [dependencies]
-//! rust-htslib = "0.38"
+//! indexmap = "1.8"
+//! noodles = { version = "0.18.0", features = ["bam", "sam", "bgzf"] }
 //! ```
 
-use std::error::Error;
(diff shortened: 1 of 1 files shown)


workflow/envs/filter_reads.yaml
```diff
@@ -2,7 +2,8 @@ channels:
   - conda-forge
 dependencies:
   - rust-script >=0.17.0
-  - rust >=1.51
+  - rust >=1.58
+  - cryptography >=36.0
   - c-compiler =1.3
   - cmake =3.21 
   - make =4.3
```
johanneskoester, line 9: Can these packages now be removed?

johanneskoester (changes requested):

workflow/envs/filter_reads.yaml
FelixMoelder: You are right. Just did a successful rerun without them.

johanneskoester (approved):

PR #113, opened 2022-02-22
fix: custom-report: Replace indexes of ANN-field by keys

FelixMoelder:
This PR updates to rbt 0.39 which allows to access the index of ANN-field entries by field names.
Also the REVEL-plot will be fixed where values being benign or malign were flipped.

### DIFF
--- a/workflow/rules/common.smk
+++ b/workflow/rules/common.smk
@@ -510,9 +510,7 @@ def get_plugin_aux(plugin, index=False):
     if plugin in config["annotations"]["vep"]["plugins"]:
         if plugin == "REVEL":
             suffix = ".tbi" if index else ""
-            return "resources/{build}_revel_scores.tsv.gz{suffix}".format(
-                build=config["ref"]["build"], suffix=suffix
-            )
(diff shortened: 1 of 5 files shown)


johanneskoester (approved):

workflow/rules/common.smk
```diff
@@ -608,6 +608,8 @@ def get_tabix_params(wildcards):
     if wildcards.format == "txt":
         return "-s 1 -b 2 -e 2"
     if wildcards.format == "tsv":
+        if config["ref"]["build"] == "GRCh37":
+            return "-f -s 1 -b 2 -e 2"
```
johanneskoester, line 612: Why the special case? At least this must be documented.

johanneskoester (changes requested):

workflow/rules/common.smk
FelixMoelder: When indexing the revel-score file the parameters are different as the file holds regions for hg38 and hg19.
Depending on which reference is being used parameters for indexing differ.

johanneskoester (approved):

PR #118, opened 2022-03-10
fix: patch picard syntax usage

dlaehnemann:
as a side-effect, this PR also turn two GitHub Action Test steps back on, that were previously skipped because all requested `.test/results` were still around from previous steps. These are now removed via a `stagein:` command. Also, we do not remove `.test/resources` in any of those `stagein:` commands any more, as that would be truly redundant testing of the resources download and processing.


tedil (approved):
Thanks for fixing!

dlaehnemann:
Thanks for merging!

And funny, that the tests simply needed a rerun to make it all the way through... :shrug:

PR #119, opened 2022-03-11
fix: picard syntax in configs and in the mark_duplicates rule

dlaehnemann:
As a side-effect, this also cleans up the `.gitignore` file a bit.

### DIFF
--- a/workflow/rules/mapping.smk
+++ b/workflow/rules/mapping.smk
@@ -40,15 +40,7 @@ rule mark_duplicates:
     log:
         "logs/picard/dedup/{sample}.log",
     params:
-        extra=lambda wc: "{c} {b} {d}".format(
-            c=config["params"]["picard"]["MarkDuplicates"],
-            b=""
-            if units.loc[wc.sample]["umis"].isnull().any()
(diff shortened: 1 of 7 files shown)


workflow/rules/mapping.smk
```diff
@@ -42,12 +42,13 @@ rule mark_duplicates:
     params:
         extra=lambda wc: "{c} {b} {d}".format(
             c=config["params"]["picard"]["MarkDuplicates"],
-            b="" if units.loc[wc.sample]["umis"].isnull().any() else "BARCODE_TAG=RX",
-            d="TAG_DUPLICATE_SET_MEMBERS=true"
+            b=""
```
tedil, line 45: I'd suggest moving that to a regular python function in common.smk

tedil (changes requested):
One small suggestion regarding the params lambda, otherwise looks fine. (I assume the .gitignore changes make sense)

workflow/rules/mapping.smk
dlaehnemann: Done. Tests are running... :runner:

tedil (approved):

PR #121, opened 2022-03-21
fix: enable specifying filter-specific vembrane options beyond the filter expression

dlaehnemann:


johanneskoester (changes requested):
Thanks! Can you please also extend the config template with an example?

workflow/schemas/config.schema.yaml
```diff
@@ -8,7 +8,10 @@ definitions:
   filterentry:
     type: object
     additionalProperties:
-      type: [string, "null"]
+      anyOf:
+        - type: string
+        - type: "null"
+        - type: object
```
johanneskoester, line 14: The object type here should be constrained to contain extra and and expression (both string).

johanneskoester (changes requested):

johanneskoester (approved):

PR #130, opened 2022-04-08
fix: automatically calculate genomic sites covered for tumor mutational burden calculation

dlaehnemann:
Instead of having to specify in the config.yaml, how much of the genome is covered, we automatically extract this from the covered regions that we already generate for focusing the variant calling steps.


workflow/rules/mutational_burden.smk
```diff
@@ -1,8 +1,19 @@
 if config["mutational_burden"]["activate"]:
 
+    rule calculate_covered_genomic_sites:
```
johanneskoester, line 3: For MTB, we need covered coding sites, don't we? Doesn't this only deliver the covered sites overall (also noncoding)?

dlaehnemann: If that is the exact definition, then yes. But then we also need to filter the mutations to only include those in coding regions. Are we doing that? I'll quickly check, currently obly on my phone...

dlaehnemann: varlociraptor does check for coding variants:

https://github.com/varlociraptor/varlociraptor/blob/master/src/estimation/mutational_burden.rs#L26-L33

So we'll need to have the intersection with coding regions here. I'll check whether Ensembl has the right BED files.

But the varlociraptor code seems a bit dangerous, should ANN field order change. Is the header parsing not there yet in rust-htslib for key-based indexing into ANN fields?

dlaehnemann: Thanks again for spotting this. I have now included a download of the reference-specific ensembl annotation, and extract all `exon`s from it that are for a `protein_coding` transcript and gene. Not sure yet, if this is really the perfect filter for the coding regions, but the resulting 80 Mb are in the right ballpark, I think (whole exome capture kits usually cover more than 60 Mb, as far as I remember).

johanneskoester (approved):

PR #148, opened 2022-05-12
feat: more informative file names

christopher-schroeder:
rename reference from "genome.fasta" to genome.{datatype}.{species}.{build}.{release}.fasta
rename sample.sorted.bam to sample.bam

### DIFF
--- a/workflow/rules/benchmarking.smk
+++ b/workflow/rules/benchmarking.smk
@@ -69,9 +69,9 @@ rule chm_eval_kit:
 
 rule chromosome_map:
     input:
-        "resources/genome.fasta.fai",
+        genome + ".fai",
     output:
-        "resources/genome.chrmap.txt",
+        "resources/" + genome_name + ".chrmap.txt",
     log:
         "logs/benchmarking/chromosome-map.log",
     conda:
@@ -83,7 +83,7 @@ rule chromosome_map:
 rule rename_chromosomes:
     input:
         bcf="results/final-calls/chm.{query}.fdr-controlled.bcf",
-        map="resources/genome.chrmap.txt",
+        map="resources/" + genome_name + ".chrmap.txt",
     output:
         "benchmarking/{query}.chr-mapped.vcf",
     params:
--- a/workflow/rules/benchmarking.smk
+++ b/workflow/rules/benchmarking.smk
@@ -69,9 +69,9 @@ rule chm_eval_kit:
 
 rule chromosome_map:
     input:
-        genome + ".fai",
+        genome_fai,
     output:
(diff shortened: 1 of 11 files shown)


christopher-schroeder:
What does the error message
fatal: Invalid symmetric difference expression master...4367e13fcb838e5fe62ffdcb5e591461d90b334e
mean?

workflow/rules/benchmarking.smk
```diff
@@ -69,9 +69,9 @@ rule chm_eval_kit:
 
 rule chromosome_map:
     input:
-        "resources/genome.fasta.fai",
+        genome_fai,
     output:
-        "resources/genome.chrmap.txt",
+        f"resources/{genome_name}.chrmap.txt",
```
johanneskoester, line 74: With Snakemake 7.8, there is no need here to have a wildcard for the genome name anymore. Changes to the genome will be picked up by Snakemake automatically since the params of the get_genome rule will change then.

johanneskoester (changes requested):
Thanks! I like the use of cram. However, to keep things simple and efficient, why don't we convert the entire pipeline to using cram (seems like the calling steps don't use it).

workflow/rules/benchmarking.smk
christopher-schroeder: I see, but its not because of the workflow behavior, it's about having the reference name in the bam file.

johanneskoester: Ah, that makes sense.

johanneskoester: I wonder whether instead of having it implicitly in the filename one should not rather make full use of the SAM spec and add AS, SP, and DS tags to the header.

johanneskoester: For assembly, species and description.

johanneskoester: This would be more machine readable than a filename that people would still need to understand.

workflow/rules/common.smk
```diff
@@ -23,6 +23,22 @@ samples = (
...
+# construct genome name
+datatype = "dna"
+species = config["ref"]["species"]
+build = config["ref"]["build"]
+release = config["ref"]["release"]
+genome_name = f"genome.{datatype}.{species}.{build}.{release}"
+genome_prefix = f"resources/{genome_name}"
+genome = f"{genome_prefix}.fasta"
+genome_fai = f"{genome}.fai"
+genome_dict = f"{genome_prefix}.dict"
```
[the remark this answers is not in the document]
johanneskoester: All fine, let's do it like this.

johanneskoester (approved):

PR #166, opened 2022-10-12
fix: correctly filter target regions

FelixMoelder:
The filter group regions filters covered regions against a set of target regions if existing.
If the target regions file exists filtering is done by `bedtools intersect` defined by `{params.filter_targets}`.
In its current implementation the covered regions and target regions are being concatenated.
This way we would end up with the whole set of target regions but also duplicate entries.

### DIFF
--- a/workflow/rules/regions.smk
+++ b/workflow/rules/regions.smk
@@ -81,7 +81,7 @@ rule filter_group_regions:
     log:
         "logs/regions/{group}.{regions_type}_regions.filtered.log",
     shell:
-        "cat {input.regions} {input.predefined} | grep -f <(head -n {params.chroms} {input.fai} | "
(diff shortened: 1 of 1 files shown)


johanneskoester (approved):

dlaehnemann:
Looking through the commits that happened while I was away, I came across this one. I think we need to revisit this, as the rule's shell code now simply doesn't use `{input.predefined}`. So instead of using cat, we should probably be doing a proper `bedtools intersect`, here. I'll look into it when I'm officially back from vacation tomorrow.

dlaehnemann:
Ah, never ming. @johanneskoester just explained it to me. The actual filtering happens via the `get_filter_targets` function result in `params.filter_targets`.

PR #171, opened 2022-11-10
fix: Fix empty regions bed file

FelixMoelder:
This PR fixes an issue where freebayes gets an empty bed-file of covered regions.
The issue occurs in case the passed target-regions file contains chromosomal positions where the chromosome is denoted with a `chr`-prefix.

### DIFF
--- a/workflow/rules/regions.smk
+++ b/workflow/rules/regions.smk
@@ -12,7 +12,7 @@ rule get_target_regions:
     conda:
         "../envs/awk_bedtools.yaml"
     shell:
-        "cat {input} | sort -k1,1 -k2,2n - | mergeBed -i - | awk '{{sub(\"^chr\",\"\", $0); print}}' > {output} 2> {log}"
+        'cat {input} | sort -k1,1 -k2,2n - | mergeBed -i - | awk \'{{sub("^chr","", $0); print}}\' > {output} 2> {log}'
 
 
 rule build_sample_regions:
--- a/workflow/rules/regions.smk
+++ b/workflow/rules/regions.smk
@@ -10,9 +10,9 @@ rule get_target_regions:
     log:
         "logs/regions/target_regions.log",
     conda:
-        "../envs/bedtools.yaml"
+        "../envs/awk_bedtools.yaml"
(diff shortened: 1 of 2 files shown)


workflow/rules/regions.smk
```diff
@@ -10,9 +10,9 @@ rule get_target_regions:
     log:
         "logs/regions/target_regions.log",
     conda:
-        "../envs/bedtools.yaml"
+        "../envs/awk_bedtools.yaml"
     shell:
-        "cat {input} | sort -k1,1 -k2,2n - | mergeBed -i - > {output} 2> {log}"
+        'cat {input} | sort -k1,1 -k2,2n - | mergeBed -i - | awk \'{{sub("^chr","", $0); print}}\' > {output} 2> {log}'
```
dlaehnemann, line 15: It feels a bit overly specific, to force this to non-`chr` Chromosome names here. But I guess we clearly only work with Ensembl reference genomes in this workflow, that have this notation, so there shouldn't be any harm in this substitution here (and I do hope there are no other species genomes in Ensembl that use `chr` as the starting part of any of their chromosomes :see_no_evil: ). Also, I guess this is something best done automagically, as simply putting a comment above the respective `config.yaml` entry `target_regions:` will probably be overlooked and as we have seen, this is a very obscure source of an error...

dlaehnemann (approved):
This generally looks good. I have more of a comment discussing my thoughts about this solution. But if you generally agree with this assessment and the comments don't trigger any further worries from your side, feel free to merge this.

workflow/rules/regions.smk
FelixMoelder: I would also argue for rely on Ensembl genomes having this kind of syntax. Making this configurable would make this more complicated for new users. If it turns out to be an issue we could still add it to the config at a later point.

christopher-schroeder: I also feel bad about forcing the non-chr. But I also think removing is way easier than adding the chr. Is there a way to do this case insensitive? I have seen "Chr" and "CHR" as well.

PR #182, opened 2022-11-23
fix: remove clinical significance duplicate

FelixMoelder:
Currently the clinical significance is rendered twice in the datavzrd report.
Once as a malformated column named `CLIN_SIG` and another correctly formatted column with identical entries.
To reduce this redundancy the corresponding entry has been removed.


dlaehnemann (approved):
Thanks!

PR #204, opened 2023-02-16
fix: update .snakemake-workflow-catalog.yml

charmoniumQ:
Add extra flags that enable this workflow to run.

See [snakemake-workflow-catalog](https://snakemake.github.io/snakemake-workflow-catalog/) and click on "standards compliant" at the top for the documentation of this YAML file.

### DIFF
--- a/.snakemake-workflow-catalog.yml
+++ b/.snakemake-workflow-catalog.yml
@@ -1,4 +1,7 @@
 usage:
+  mandatory-flags:
+    flags: "--directory .test --configfile .test/config-simple/config.yaml"
+    desc: Use the `.test` directory and the sample configuration
   software-stack-deployment:
     conda: true
   report: true


johanneskoester:
The mandatory flags are actually not meant to provide a hint for how to run on the test data. The test data isn't even present when you run through the deployment steps presented by the workflow catalog. Also, this would be misleading as people could think that you always need to use --directory to run the workflow.

johanneskoester:
Thinking about it, you nevertheless have a case here. It does not hurt to have a mechanism in the catalog config to display how to run an example run with test data. I am happy to take a proposal for that within the workflow catalog repository, and then we could fill this out here. Maybe:
```
example-test: |
  wget <url-to-test-data>
  unzip ...
  cd ...
  snakemake ...
```

dlaehnemann:
Maybe it would also make sense to automate the creation of such a test scenario from an existing test in the `.test/` directory. For example if there is one called `.test/minimal/`. But maybe this is also making it too complicated to get it off the ground...

charmoniumQ:
My motivation for this was because I am working on automatically running snakemake workflows (no manual intervention). Currently I am going over repositories in [snakemake workflow catalog][1], applying options from the `.snakemake-workflow-catalog.yml`, and from github CI `./github/workflows/*.yaml`. [Here][2] is my code. Despite a lot of trying stuff, I can only get ~5% of the Snakemake workflows to run.

If you know anyone who would be interested in using or contributing to a standard for specifying how to run Snakemake workflows, or even just using the data, let me know <mailto:grayson5@illinois.edu>.

[1]: https://snakemake.github.io/snakemake-workflow-catalog/
[2]: https://github.com/charmoniumQ/wf-reg-test/blob/main/wf_reg_test/engines.py#L116

PR #213, opened 2023-03-02
feat: handle umis

FelixMoelder:
Until now UMIs where only supported by adding a fastq file containing the UMI of each read.
Often UMIs do not exists as separate fastq records but as part of the read sequences.
To handle UMIs properly information about them is now stored in two additional columns in the samplesheet.
- `umi_read`: Defines whether UMIs are part of records in fq1 or fq2.
- `umi_read_structure`: The template of the read defining the position of UMIs in records (see https://github.com/fulcrumgenomics/fgbio/wiki/Read-Structures)

Handling UMIs is optional. In case the `umi_read` column is missing or left empty UMIs will not be annotated for duplicate marking or consensus read calculation.

### DIFF
--- a/workflow/rules/mapping.smk
+++ b/workflow/rules/mapping.smk
@@ -15,24 +15,39 @@ rule map_reads:
         "v1.10.0/bio/bwa/mem"
 
 
+rule merge_untrimmed_fastqs:
+    input:
+        get_untrimmed_fastqs,
+    output:
+        temp("results/untrimmed/{sample}_{read}.fastq.gz"),
+    log:
+        "logs/merge-fastqs/untrimemd/{sample}_{read}.log",
+    wildcard_constraints:
+        read="fq1|fq2",
+    shell:
+        "cat {input} > {output} 2> {log}"
+
+
 rule annotate_umis:
     input:
         bam="results/mapped/{sample}.bam",
-        umi=lambda wc: units.loc[wc.sample]["umis"][0],
+        umi=get_umi_fastq,
     output:
         temp("results/mapped/{sample}.annotated.bam"),
+    params:
+        extra=get_umi_read_structure,
     resources:
         mem_gb="10",
     log:
         "logs/fgbio/annotate_bam/{sample}.log",
     wrapper:
-        "v1.2.0/bio/fgbio/annotatebamwithumis"
+        "v1.23.4/bio/fgbio/annotatebamwithumis"
 
 
 rule mark_duplicates:
     input:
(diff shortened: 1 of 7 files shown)


johanneskoester (commented):
We should discuss how this is joined with the separate umi fastq case. How is that configured now?

workflow/rules/mapping.smk
```diff
@@ -15,24 +15,39 @@ rule map_reads:
...
     wrapper:
-        "v1.2.0/bio/fgbio/annotatebamwithumis"
+        "v1.23.4/bio/fgbio/annotatebamwithumis"
 
 
 rule mark_duplicates:
     input:
         lambda wc: "results/mapped/{sample}.bam"
-        if units.loc[wc.sample, "umis"].isnull().any()
+        if pd.isna(samples.loc[wc.sample, "umi_read"])
```
johanneskoester, line 50: Rather have a helper function for this: `sample_has_umis(wc.sample)`

FelixMoelder:
> We should discuss how this is joined with the separate umi fastq case. How is that configured now?

This still works. In case we have a separate fastq file with umis one can just define that file and set the read structure to `+M` which defines the whole sequence in the fastq is the UMI.

johanneskoester:
> > We should discuss how this is joined with the separate umi fastq case. How is that configured now?
> 
> This still works. In case we have a separate fastq file with umis one can just define that file and set the read structure to `+M` which defines the whole sequence in the fastq being the UMI.

Can you update config/README.md to describe all ways to configure UMIs please?

config/README.md
```diff
@@ -40,3 +40,11 @@ In case of different panels, primers have to be set panel-wise in a seperate tsv
 For each panel the following columns need to be set: `panel`, `fa1` and `fa2` (optional).
 Additionally, for each sample the corresponding panel must be defined in `samples.tsv` (column `panel`).
 For single primer trimming only, the first entry in the config (respective in the tsv file) needs to be defined.
+
+# Annotating UMIS
+
+For annotating UMIs two additional columns in `sample.tsv` must be set:
+* umi_read: This can be either `fq1` or `fq2` if the UMIs are part of one of the two. Alternatively, a fastq file containing UMIs can be set.
```
johanneskoester, line 47:
```suggestion
* umi_read: This can be either `fq1` or `fq2` if the UMIs are part of one of the two. Alternatively, the path to an additional fastq file containing just the UMI of each fragment in fq1 and fq2 (with the same read names) can be set.
```

Issue #216, opened 2023-03-11
modify the rule filter_offtarget_variants or fix_delly_calls


user-tq:
get erro
```
[E::idx_find_and_load] Could not retrieve index file for 'results/candidate-calls/2022NC.delly.no_bnds.bcf'
Failed to read from results/candidate-calls/2022NC.delly.no_bnds.bcf: could not load index
```
![image](https://user-images.githubusercontent.com/72878144/224495914-b33e40f7-8c3f-4e71-9893-e4e7721add8a.png)
Maybe it can be changed to
```
rule fix_delly_calls:
    input:
        "results/candidate-calls/{group}.delly.bcf",
    output:
        "results/candidate-calls/{group}.delly.no_bnds.bcf",
    log:
        "logs/fix_delly_calls/{group}.log",
    conda:
        "../envs/bcftools.yaml"
    shell:
        """bcftools view -e 'INFO/SVTYPE="BND"' {input} -Ob > {output} 2> {log} && bcftools index   {output} """
```

dlaehnemann:
I think, contrary to the `[E::` at the start of the message, this is not actually an error, but only a warning, and `bcftools filter` works without the index file. Or are you seeing this rule fail because of this error?

And should this really be a hard failure, we would need to fix this in the [`bcftools filter` wrapper](https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/bcftools/filter.html), requiring the bcf index file as an input there. Here, in the workflow, this would then mean adjusting the input to `rule filter_offtarget_variants` to also include the index file, which should then automatically be created by the `rule bcftools_index` in `workflow/rules/utils.smk`:

https://github.com/snakemake-workflows/dna-seq-varlociraptor/blob/9d271ef3bf235f2dd63616d08d3b4e3f22cbe019/workflow/rules/utils.smk#L1-L11

dlaehnemann:
Do you have any follow-up info here, or can we close this? Namely, is this really just an irritating warning that sounds like an error, but does not actually stop/fail the workflow.

user-tq:
> Do you have any follow-up info here, or can we close this? Namely, is this really just an irritating warning that sounds like an error, but does not actually stop/fail the workflow.

![image](https://user-images.githubusercontent.com/72878144/224974169-e48f344f-b6bb-43bb-90ff-cc46be4fa96d.png)
[...]

PR #220, opened 2023-03-14
fix: require bcf index as input to rule filter_offtarget_variants

dlaehnemann:
This is the quick fix for #216 . For the more thorough fix in snakemake-wrapper-utils, we should keep that issue open and I'll document what to do over there.


FelixMoelder (approved):
Looks good to me! :)

dlaehnemann:
While you're at it, want to also review the more thorough fix:
https://github.com/snakemake/snakemake-wrapper-utils/pull/26

PR #222, opened 2023-03-16
feat: fusion calling

FelixMoelder:
These changes introduce fusion calling in addition the variant calling.
There for arriba is applied for calling variants. As vep does not provide any helpful information it is skipped for fusion calls.
Instead annotation is performed by additional scripts that come with the arriba package.
To define a what kind of analysis shall be performed for a sample an additional column named `datatype` has been added to the samplesheet allowing either `rna` or `dna`. As file name extensions within the pipeline are separated into `variants` and `fusions` we probably should unify this as the former one might also be too generic.
For fusion calls a separte datavzrd report will be created adding a new category showing up in the final snakemake report.


FelixMoelder:
To get this running we need https://github.com/suhrig/arriba/pull/191 being merged and a new release of arriba.

workflow/rules/common.smk
```diff
@@ -539,17 +597,22 @@ def get_selected_annotations():
 
 
 def get_annotated_bcf(wildcards):
```
FelixMoelder, line 599: We could add `datatype = "variants" if wildcards.analysis == "variants" else "arriba"`.
While it would be a bit unintuitive replacing `fusions` as the default datazype by the caller `arriba` this would skip calling `bcftools concat` where `delly` and `freebayes` calls are concatenated for variant calling.
For fusion calls there is only one caller and this step could be skipped at the expanse of a bit confusing datatype assignment.

johanneskoester: not a fan of that. I'd rather skip it in a different way.

johanneskoester (changes requested):

config/README.md
```diff
@@ -4,11 +4,12 @@ To configure this workflow, modify ``config/config.yaml`` according to your need
...
 * `alias`es represent the name of the sample within its group. They are meant to be some abstract description of the sample type to be used in the [Calling scenario](#calling-scenario), and should thus be used consistently across groups. A classic example would be a combination of the `tumor` and `normal` aliases.
 * The `platform` column needs to contain the used sequencing plaform (one of 'CAPILLARY', 'LS454', 'ILLUMINA', 'SOLID', 'HELICOS', 'IONTORRENT', 'ONT', 'PACBIO’).
 * The same `sample_name` entry can be used multiple times within a `samples.tsv` sample sheet, with only the value in the `group` column differing between repeated rows. This way, you can use the same sample for variant calling in different groups, for example if you use a panel of normal samples when you don't have matched normal samples for tumor variant calling.
+* The `datatype` column specifies what kind of data each sample corresponds to. This can either be `fusions` or `variants`. For fusion samples the workflow performs fusion calling while variant samples are utilized for calling variants.
```
johanneskoester, line 12: Mhm, this we should discuss, maybe tomorrow.

johanneskoester (changes requested):

config/config.yaml
```diff
@@ -123,6 +123,7 @@ calling:
       # only alphanumerics and underscores
       # ("somatic" below is just an example and can be modified as needed).
```
johanneskoester, line 124:
```suggestion
      # ("some_id" below is just an example and can be modified as needed).
```

config/config.yaml
```diff
@@ -123,6 +123,7 @@ calling:
...
       some_id:
+        types: ["variants", "fusions"]
```
johanneskoester, line 126: Add a comment above this line explaning what this is about.

workflow/report/workflow.rst
```diff
@@ -1,7 +1,7 @@
 This workflow generates annotated variant calls that can be viewed in interactive reports, showing all evidence levels provided by Varlociraptor_.
 Adapters were removed with Cutadapt_. Reads were mapped with `BWA MEM`_, PCR and optical duplicates were removed with Picard_.
 Candidate variant discovery was performed with Freebayes_ and Delly_. Statisticall assessment of variants was conducted with Varlociraptor_.
-Variant calling results, sorted by type, event, and impact can be found under `Variant calls`_.
+Fusion resp. Variant calling results, sorted by type, event, and impact can be found under Fusion/Variant calls.
```
johanneskoester, line 4:
```suggestion
Fusion resp. variant calling results, sorted by type, event, and impact can be found under Fusion/Variant calls.
```

workflow/rules/datavzrd.smk
```diff
@@ -86,16 +128,45 @@ rule datavzrd_variants_calls:
...
         report(
-            directory("results/datavzrd-report/{batch}.{event}.fdr-controlled"),
+            directory(
+                "results/datavzrd-report/{batch}.{event}.variants.fdr-controlled"
+            ),
             htmlindex="index.html",
             caption="../report/calls.rst",
             category="Variant calls",
-            labels=get_datavzrd_report_labels,
+            labels=lambda wc: get_datavzrd_report_labels(wc),
```
johanneskoester, line 142: superfluous lambda

workflow/rules/datavzrd.smk
```diff
@@ -86,16 +128,45 @@ rule datavzrd_variants_calls:
...
+        config="resources/datavzrd/{batch}.{event}.fusions.datavzrd.yaml",
+    output:
+        report(
+            directory(
+                "results/datavzrd-report/{batch}.{event}.fusions.fdr-controlled"
+            ),
+            htmlindex="index.html",
+            caption="../report/calls.rst",
+            category="Fusion calls",
+            labels=lambda wc: get_datavzrd_report_labels(wc),
```
johanneskoester, line 169: superfluous lambda

workflow/rules/datavzrd.smk
```diff
@@ -71,7 +84,36 @@ rule render_datavzrd_config:
...
+
+rule render_datavzrd_fusions_config:
+    input:
+        template=workflow.source_path(
+            "../resources/datavzrd/fusion-calls-template.datavzrd.yaml"
+        ),
+    output:
+        "resources/datavzrd/{batch}.{event}.fusions.datavzrd.yaml",
+    params:
+        groups=lambda wc: get_report_batch(wc, "fusions"),
```
johanneskoester, line 100: Wrap this so that the lambda is not necessary. Let get_report_batch take the type as input and return a function that takes wildcards as input.

workflow/scripts/convert_fusions_to_vcf.sh
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
```
johanneskoester, line 1: Does this script come from arriba sources? Please cite the full URL here.

johanneskoester: And is the license compatible?

johanneskoester (changes requested):

workflow/scripts/convert_fusions_to_vcf.sh
FelixMoelder: The license of arriba is also MIT and therefore compatible. :)

workflow/scripts/convert_fusions_to_vcf.sh
```diff
@@ -0,0 +1,94 @@
...
+
+# this script is part of arriba (https://github.com/suhrig/arriba) and published under the MIT Expat license.
+# source: https://github.com/suhrig/arriba/blob/master/scripts/convert_fusions_to_vcf.sh
```
johanneskoester, line 4: Please use the commit you have taken this from instead of master branch in the URL..

johanneskoester (changes requested):

johanneskoester (approved):

PR #225, opened 2023-04-24
feat: Support new varlociraptor 8.0 observations

FelixMoelder:
This adds support for varlociraptor 8.0 short observations and regular observation fields.
In case of short observations (`SAOBS` and `SROBS`) ref and alt alleles columns are being joined in order to be processable by datavzrd.
For the common `OBS`-field the edit distance is now being considered.

### DIFF
--- a/workflow/scripts/split-call-tables.py
+++ b/workflow/scripts/split-call-tables.py
@@ -125,37 +125,16 @@ def cleanup_dataframe(df):
     return df
 
 
-def modify_observations(value, modifier):
-    regex = re.compile("([0-9]+)(E|B|P|S|V|e|b|p|s|v)")
-    observations = ""
-    m = re.findall(regex, value)
-    for count, score in m:
-        if score != "E":
-            observations += count + modifier + score
-        else:
-            observations += count + score
-    return observations
-
-
-def modify_observations(value, modifier):
-    regex = re.compile("([0-9]+)(E|B|P|S|V|e|b|p|s|v)")
-    observations = ""
-    m = re.findall(regex, value)
-    for count, score in m:
-        if score != "E":
-            observations += count + modifier + score
-        else:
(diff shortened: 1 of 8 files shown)


workflow/scripts/split-call-tables.py
```diff
@@ -124,6 +125,46 @@ def cleanup_dataframe(df):
...
+def modify_observations(value, modifier):
+    regex = re.compile("([0-9]+)(E|B|P|S|V|e|b|p|s|v)")
+    observations = ""
+    m = re.findall(regex, value)
+    for count, score in m:
+        if score != "E":
+            observations += count + modifier + score
+        else:
+            observations += count + score
+    return observations
```
johanneskoester, line 149: Two times the same function?

workflow/scripts/split-call-tables.py
```diff
@@ -124,6 +125,46 @@ def cleanup_dataframe(df):
...
+
+
+def join_short_obs(df, samples):
+    for sample in samples:
+        df[f"{sample}: short observations"] = ""
+        for allele, abbrev in [("ref", "R"), ("alt", "A")]:
+            df[f"{sample}: short observations"] += df[
+                f"{sample}: short {allele} observations"
+            ].apply(modify_observations, args=(abbrev))
```
johanneskoester, line 158: I think I would prefer to keep the observations unmodified, and just concatenate them with a comma or so. The reason: we want to migrate the JS parsing into an npm library, and there it has to deal with the original format for SROBS and SAOBS.

johanneskoester (changes requested):

workflow/scripts/split-call-tables.py
FelixMoelder: I changed it the way you suggested. :)

PR #244, opened 2023-06-13
fix: update varlociraptor

FelixMoelder:
Update for varlociraptor as it fixes an issue occuring when a record does not have an `ANN` field.


workflow/scripts/split-call-tables.py
```diff
@@ -92,6 +92,9 @@ def order_impact(df):
 
 
 def sort_calls(df):
+    sorting = snakemake.params.sorting
+    if "revel" not in df.columns:
+        sorting.remove("revel")
```
FelixMoelder, line 97: This is the relevant change. All other changes are only reformatted.

PR #250, opened 2023-06-29
fix: hide binned vaf

FelixMoelder:
Hiding the `binned max vaf` column in the datavzrd report as it is only required for sorting.


FelixMoelder:
Also removed the varsome link for gene oncoprints. For creating the link ref and alt alleles are required. As those are not part of the oncoprint table the datavzrd report return a type error.

johanneskoester (approved):

PR #251, opened 2023-07-26
feat!: make annotations for candidate variant filtering configurable

dlaehnemann:
We have a use case, where we want to use gnomAD population frequencies of variants to filter out common variants BEFORE running varlociraptor. For this, we need to be able to customize the vep flags (and potentially plugins in the future) used for annotating candidate calls before candidate call filtering. This PR introduces this.

### DIFF
--- a/workflow/schemas/config.schema.yaml
+++ b/workflow/schemas/config.schema.yaml
@@ -255,6 +255,9 @@ properties:
             required:
               - params
               - plugins
+        required:
+          - candidate_calls
+          - final_calls
     required:
       - vep
       - vcfs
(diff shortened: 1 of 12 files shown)


johanneskoester (approved):

workflow/schemas/config.schema.yaml
```diff
@@ -233,15 +233,28 @@ properties:
...
+            properties:
+              params:
+                type: string
+              plugins:
+                type: array
+                items:
+                  type: string
+            required:
+              - params
+              - plugins
```
johanneskoester, line 257: I think this now lacks a "required"-section for the two cases.

johanneskoester (changes requested):

workflow/schemas/config.schema.yaml
dlaehnemann: It does. Added them just now. Thanks for catching this!

johanneskoester (approved):

PR #255, opened 2023-08-09
feat: update envs and wrappers

FelixMoelder:


dlaehnemann:
I found another GitHub Action to free space from (hopefully) unused stuff. And I just went ahead to try it. Here's the docs:
https://github.com/marketplace/actions/free-disk-space-ubuntu

dlaehnemann:
So. this seems to run through, now. So I'd say this is ready for review.

One quick question from my side: Is it intentional that the test case on actual SRA data is only a dry-run? Is this to save time? See:
https://github.com/snakemake-workflows/dna-seq-varlociraptor/blob/f112012b5485a2102dadf4690779f3bfd0ca1859/.github/workflows/main.yml#L132

PR #256, opened 2023-08-09
chore: try large runner

johanneskoester:


johanneskoester:
Does not work. Large runners are apparently not part of our always-free github teams coupon.

PR #258, opened 2023-08-09
fix: allow using one sample in multiple groups

dlaehnemann:


dlaehnemann:
This is ready for review, now. Basically, all I had to do to allow for the same sample to be used in multiple groups (i.e. in multiple rows in the `samples.tsv`, which only differ by the `group` specification), was to make sure that `platform` extraction from the sample sheet always returns just a single `str` value. The rest of the changes were all for getting the CI test to cover this.

johanneskoester (approved):

johanneskoester (changes requested):
Can you extend the `config/README.md` such that it explicitly explains that this is allowed?

dlaehnemann:
Yes, of course. I had that on my list, but forgot about it during debugging of the CI... Thanks for catching this!

dlaehnemann:
Did a bit of a cleanup of the `config/README.md` file. I had also forgotten to update it for the `UMI` changes and I though that some of the wording and structure could be improved. Feedback welcome!

johanneskoester (approved):

PR #263, opened 2023-08-24
fix: general fixes and optimizations

FelixMoelder:
Since updating all integrated wrappers the workflow fails at certain points.
Besides we write a lot of intermediate fastq and bam files without deleting them. For large samples this might end up in over a hundred gigabytes of additional disk storage. I therefore marked all of these intermediate results as temporary only keeping merged fastqs and recalibrated bam files.

List of changes:
- updated datavzrd to latest version
- updated `vep annotate` wrapper as it still used `vep v109` while `vep cache` already already used `v110`
- set memory for picard to 3gb
  - the faq recommand to use 1GB per 250.000 reads in memory with length of 100bp. Per default there are 500.000reads in memory and I assumed a read length of 150bp ending up with 3GB. In addition broad uses internally at least 2GB (https://broadinstitute.github.io/picard/faq.html)
- downgrade `gatk baserecalibrator` wrapper as baserecalibration fails.
  - the error seems to be caused by gatk as the error says that the known variants gtf file can not found which is previously copied to a temporary directory by the tool it self. Considering the documentation there has nothing changed how gatk is run so this seems to be a bug.
- updated `ensembl-variation` wrapper as the previous version failed due to downloaded index files

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -22,7 +22,7 @@ ref:
   # Ensembl species name
   species: homo_sapiens
   # Ensembl release
-  release: 109
+  release: 110
   # Genome build
   build: GRCh38
   # Optionally, instead of downloading the whole reference from Ensembl via the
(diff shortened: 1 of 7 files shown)


FelixMoelder:
This is ready to review. Still, before merging we should update datavzrd once available as the was a rendering error fixed.

johanneskoester (approved):

PR #270, opened 2023-10-26
feat: gene coverage report

FelixMoelder:
This adds an additional datavzrd report showing the gene mean gene coverages per sample group-wise.

While the annotation file previously was a gzip file we now directly output a plain gtf file.
This must be done for fusion calling anyway as zipped files are not supported.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -57,6 +57,9 @@ mutational_burden:
     - somatic_tumor_medium
     - somatic_tumor_high
 
+gene_coverage:
+  min_avg_coverage: 5
+
 # printing of variants in interactive tables
 report:
   # if stratification is deactivated, one report for all
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -57,6 +57,9 @@ mutational_burden:
     - somatic_tumor_medium
     - somatic_tumor_high
 
+# Sets the minimum average coverage for each gene.
+# Genes with lower average coverage will not be concidered in gene coverage datavzrd report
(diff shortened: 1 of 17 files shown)


config/config.yaml
```diff
@@ -57,6 +57,9 @@ mutational_burden:
     - somatic_tumor_medium
     - somatic_tumor_high
 
+gene_coverage:
+  min_avg_coverage: 5
```
johanneskoester, line 61: Please add explaining comments.

workflow/rules/ref.smk
```diff
@@ -59,9 +59,9 @@ rule get_known_variants:
         "v2.5.0/bio/reference/ensembl-variation"
 
 
-rule get_annotation_gz:
+rule get_annotation:
     output:
-        "resources/annotation.gtf.gz",
+        "resources/annotation.gtf",
```
johanneskoester, line 64: I would not do this here. Am I right that this is only needed for the scripts that postprocess this? In the scripts, you should be able to just read the file as gzip with python: https://docs.python.org/3/library/gzip.html

johanneskoester (changes requested):

workflow/rules/ref.smk
FelixMoelder: This is indeed not needed for this PR but for the fusion calling PR where the star aligner also requires the annotations. As star does not support gzip-files I decided to already switch to a plain gtf file in this PR as I would just require to overhaul the python script in the other PR.

johanneskoester (approved):

PR #274, opened 2023-11-16
feat: add option to allow mapping to pangenome

huzuner:
This PR allows an optional pangenome alignment with `vg giraffe`. 
The following additions/changes are made:

- A new option `pangenome` is added to the config.
- A rule to download the vg pangenome index is added into `ref.smk`.
-  Six other rules for aligning reads to the pangenome and preprocessing the resulting bam were added, into `mapping.smk`.
- All input functions are updated to get pangenome aligned bam upon activation of `pangenome` in the config file.

## Summary by CodeRabbit

- **New Features**
	- Introduced pangenome configuration options for genomic analysis, allowing for future enhancements in variant calling.
	- Added new workflow rules for pangenome data processing, including haplotype downloading, chromosome renaming, and improved read mapping with the VG tool.
	- Expanded event types in variant calling processes.
	- Implemented a script for renaming contigs in VCF files based on user-defined expressions.

- **Bug Fixes**
	- Improved flexibility in aligner selection for RNA samples.

- **Chores**
	- Updated dependency versions for `varlociraptor` and `bcftools` to ensure compatibility.
	- Created a new Conda environment configuration for `pysam`.


huzuner:
Bug with vg: alignments with deletions on ends: https://github.com/vgteam/vg/issues/4204#issuecomment-1937889331. They fixed the issue and the new release is coming at Feb. 26.

workflow/rules/common.smk
```diff
@@ -239,7 +239,10 @@ def get_recalibrate_quality_input(wildcards, bai=False):
     elif is_activated("remove_duplicates"):
         return "results/dedup/{{sample}}.{ext}".format(ext=ext)
     else:
-        return "results/mapped/bwa/{{sample}}.{ext}".format(ext=ext)
+        if is_activated("ref/pangenome"):
+            return "results/mapped/vg/{{sample}}_rg_added.{ext}".format(ext=ext)
+        else:
+            return "results/mapped/bwa/{{sample}}.{ext}".format(ext=ext)
```
johanneskoester, line 245: Use f-string in both cases, use `{mapper}` or something like that in order to not mostly repeat the f-string.

workflow/rules/common.smk
```diff
@@ -409,15 +419,29 @@ def get_consensus_input(wildcards):
...
     else:
-        aligner = "star" if get_sample_datatype(wildcards.sample) == "rna" else "bwa"
+        if get_sample_datatype(wildcards.sample) == "rna":
+            aligner = "star"
+        elif get_sample_datatype(wildcards.sample) == "dna" & is_activated(
+            "ref/pangenome"
+        ):
+            aligner = "vg"
+        else:
+            aligner = "bwa"
```
johanneskoester, line 429: Same code as above, create new function `get_aligner(wildcards)` and use in both blocks.

johanneskoester: and below as well

workflow/rules/filtering.smk
```diff
@@ -19,6 +19,7 @@ rule filter_candidates_by_annotation:
 rule filter_by_annotation:
     input:
         bcf=get_annotated_bcf,
+        csi=lambda wc: get_annotated_bcf(wc, index=True),
```
johanneskoester, line 22:
```suggestion
        csi=partial(get_annotated_bcf, index=True),
```

workflow/rules/mapping.smk
```diff
@@ -214,3 +214,126 @@ rule apply_bqsr:
         java_opts="",  # optional
     wrapper:
         "v2.3.2/bio/gatk/applybqsr"
+
+
+rule map_reads_vg_giraffe:
+    input:
+        reads=get_map_reads_input,
+        idx="resources/pangenome/hprc-v1.0-mc-grch38.xg",
```
johanneskoester, line 222: This should not be hardcoded. What if a different pangenome is used? The pangenome index should be configurable via the config.

workflow/rules/mapping.smk
```diff
@@ -214,3 +214,126 @@ rule apply_bqsr:
...
+        "results/mapped/vg/{sample}_sorted.bai",
+    output:
+        bam="results/mapped/vg/{sample}_extracted.bam",
+        idx="results/mapped/vg/{sample}_extracted.bai",
+    log:
+        "logs/samtools_view_primary_chr/{sample}.log",
+    benchmark:
+        "benchmarks/samtools_view_primary_chr/{sample}.tsv"
+    params:
+        region="GRCh38.chr1 GRCh38.chr2 GRCh38.chr3 GRCh38.chr4 GRCh38.chr5 GRCh38.chr6 GRCh38.chr7 GRCh38.chr8 GRCh38.chr9 GRCh38.chr10 GRCh38.chr11 GRCh38.chr12 GRCh38.chr13 GRCh38.chr14 GRCh38.chr15 GRCh38.chr16 GRCh38.chr17 GRCh38.chr18 GRCh38.chr19 GRCh38.chr20 GRCh38.chr21 GRCh38.chr22 GRCh38.chrX GRCh38.chrY GRCh38.chrM",
```
johanneskoester, line 267: This should not be hardcoded. Workflow has to work for all species in principle (as long as there is a pangenome index). We have a selection for the first n chromosomes already implemented in the workflow (at the level of the variants). Isn't that enough?

workflow/rules/mapping.smk
```diff
@@ -214,3 +214,126 @@ rule apply_bqsr:
...
+    output:
+        "results/mapped/vg/{sample}.bam",
+    log:
+        "logs/picard/add_rg/{sample}.log",
+    params:
+        extra="--RGLB lib1 --RGPL illumina --RGPU {sample} --RGSM {sample}",
+    resources:
+        mem_mb=60000,
+    wrapper:
+        "v2.3.2/bio/picard/addorreplacereadgroups"
```
johanneskoester, line 325: Are you sure that it is impossible to define the readgroup directly when calling vg giraffe?

workflow/rules/ref.smk
```diff
@@ -146,3 +146,18 @@ rule get_vep_plugins:
...
         "v3.3.5/bio/vep/plugins"
+
+
+# github link to the vg index for "xg" extension with the corresponding commit ID:
+# https://github.com/human-pangenomics/hpp_pangenome_resources/tree/201b532d11e8ba75a97db5005b98a780f42feeb6
+
+
+rule get_vg_pangenome:
+    output:
+        "resources/pangenome/hprc-v1.0-mc-grch38.xg",
```
johanneskoester, line 157: Rather name this analogously to the bwa index output, hence:
```suggestion
        f"{genome}.xg",
```

johanneskoester (changes requested):

config/config.yaml
```diff
@@ -139,8 +142,9 @@ calling:
           # Add varlociraptor events to aggregated over.
           # The probability for the union of these events is used for controlling
           # the FDR.
-          - somatic_tumor_high
-          - somatic_tumor_medium
```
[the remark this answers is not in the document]
huzuner: updated to use default workflow config and scenario.

workflow/rules/mapping.smk
huzuner: With vg, it is possible to define read groups for sample names but not for platforms. BaseRecalibrator later complains about it: 
```
org.broadinstitute.hellbender.exceptions.UserException$MalformedRead: Read SRR702070.29886333 1:13000-13100 is malformed: The input .bam file contains reads with no platform information. First observed at read with name = SRR702070.29886333
```

config/config.yaml
```diff
@@ -112,7 +115,7 @@ calling:
     gene_list_filter:
       aux-files:
         super_interesting_genes: "config/super_interesting_genes.tsv"
-      expression: "ANN['GENE'] in AUX['super_interesting_genes']"
```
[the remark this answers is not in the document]
FelixMoelder: Actually the column name is `SYMBOL` and never was `GENE`. We probably never validated this config for correctness.

workflow/rules/ref.smk
huzuner: renamed and link to download was added to config for exemplary purposes.

workflow/rules/mapping.smk
huzuner: Do you refer to a function that looks in the header, finds the names of first n chromosomes from `config['ref']['n_chromosomes']`  and that is then used here to extract only records belonging to them? Wouldn't that be the only way other than hardcoding them? As far as I checked, that kind of implementation doesn't exist in the workflow, and the vg index is in `xg` format that cannot be easily subset like a fasta. Can you eloborate further on how we can achieve what you mean?

workflow/rules/mapping.smk
huzuner: I added `config["link_to_pangenome_idx"]` to use in the download rule and download output path to common_smk.

johanneskoester (changes requested):

workflow/rules/mapping.smk
```diff
@@ -214,3 +214,126 @@ rule apply_bqsr:
...
+        "logs/mapped/vg/{sample}.log",
+    benchmark:
+        "benchmarks/vg_giraffe/{sample}.tsv"
+    conda:
+        "../envs/vg.yaml"
+    threads: 40
+    params:
+        lambda wc, input: " -f ".join(input.reads),  # potential issue: in case of single end reads, get map_reads_input() returns a string and join() could create a problem.
+    shell:
+        "vg giraffe -x {input.idx} -f {params} --output-format BAM -t {threads}  > {output} 2> {log}"  #read groups for samples can be added with: --sample {wildcards.sample} --read-group {wildcards.sample}
```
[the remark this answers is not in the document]
huzuner: Suggestion applied.

workflow/envs/vg.yaml
```diff
@@ -0,0 +1,5 @@
+name: vg
+channels:
+   - bioconda
+dependencies:
+   - vg
```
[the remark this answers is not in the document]
huzuner: Suggestion applied.

huzuner:
Regarding the error in tests and the latest change on Snakemake, should this line use conda?:https://github.com/snakemake-workflows/dna-seq-varlociraptor/blob/6c5cc97f9c033a7422b41c6108c2a4e608a55b0b/.github/workflows/main.yml#L38

workflow/rules/common.smk
```diff
@@ -30,6 +30,9 @@ genome_prefix = f"resources/{genome_name}"
 genome = f"{genome_prefix}.fasta"
 genome_fai = f"{genome}.fai"
 genome_dict = f"{genome_prefix}.dict"
+# in case pangenome is used
+pangenome_path = "resources/pangenome/vg_index.xg"
+pangenome = f"{pangenome_path}"
```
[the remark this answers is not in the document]
huzuner: Pangenome index path is now being called from config.

FelixMoelder:
When running the workflow with UMI annotation `fgbio AnnotateBamsWithUmis` fails as the reads needs to be name sorted.
For bwa this is implemented in the wrapper. For `vg grraffe` we should create a wrapper providing the same sorting options (coordinate/querysorted, samtools/fgbio) as for bwa.

FelixMoelder:
`vg giraffe` is creating several files next to the reference graph like a distance index (pangenome.homo_sapiens.GRCh38.dist).
When running the workflow with multiple samples this might become if both processes write to the same file.
I did not find a possibility to create these files before mapping.

FelixMoelder:
Notes to myself to continue after vacation:
- parallel mapping of reads by `vg giraffe` not fully supported, yet
   - if only `xg`-index is provided `.dist`-, `.min`- and `gbz`-files will be created on first execution (could lead to race conditions/corrupted files when multiple samples are mapped at the same time)
   - `.dist`- and `.min`-files are being downloaded before mapping
   - `gbz`-file is not available for download (can it be created locally?)
- currently only hprc-v1.0 is supported
   - hprc-v1.1 graph is available as `bgz`-file and allows creating all additional files first
   - hprc-v1.1 graph also contains CHM13, seems like reads are mapped against GRCh38 and CHM13 leading to downstream errors
   - hprc-v1.1 chromosomes are annotated differently in the bam header (needs differentiation between v1.0 and v1.1)
- `varlociraptor preprocess` may fail as `vg giraffe` creates CIGARs containing leading deletions

huzuner:
Trailing deletions in the CIGAR was a bug in `vg giraffe` before in the previous releases but should not be an issue any more in the current version we use. They fixed that before and if it happens again we could create an issue in their repo.

FelixMoelder:
> Trailing deletions in the CIGAR was a bug in `vg giraffe` before in the previous releases but should not be an issue any more in the current version we use. They fixed that before and if it happens again we could create an issue in their repo.

You are right, it should be fixed since 1.55 but I was using 1.59. So it appears to be still an issue.

workflow/rules/mapping.smk
```diff
@@ -15,13 +15,102 @@ rule map_reads:
...
+    input:
+        "results/mapped/vg/{sample}_reheadered.bam",
+    output:
+        "results/mapped/vg/{sample}.bam",
+    log:
+        "logs/picard/add_rg/{sample}.log",
+    params:
+        extra="--RGLB lib1 --RGPL ILLUMINA --RGPU {sample} --RGSM {sample} --RGID {sample}",
+    resources:
+        mem_mb=60000,
```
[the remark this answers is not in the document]
huzuner: Sorry I just see this mention now!
To answer your question, honestly, I don't remember. We can remove it and just put a warning note that it might throw memory errors to just use as a hint in the future.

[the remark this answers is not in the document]
FelixMoelder: I will just set this to 1024MB and see if it works for real data

workflow/rules/ref.smk
```diff
@@ -146,3 +146,57 @@ rule get_vep_plugins:
...
+    output:
+        temp("resources/chrom_renames.tsv"),
+    log:
+        "logs/pangenome/chrom_replacement.log",
+    conda:
+        "../envs/bcftools.yaml"
+    shell:
+        """
+        bcftools query -f '%CHROM\n' {input} | uniq | awk '{{if ($1 == "chrM") print $1 "\tMT"; else print $1 "\t" substr($1, 4)}}' > {output} 2> {log}
+        """
```
johanneskoester, line 174: Make this configurable
```yaml
rename-expressions:
  - contig[3:]
  - "MT" if contig == "M" else contig
```

johanneskoester (changes requested):

workflow/rules/ref.smk
```diff
@@ -146,3 +146,57 @@ rule get_vep_plugins:
...
+        temp(f"{pangenome_prefix}.vcf.gz"),
+    params:
+        url=config["ref"]["pangenome"]["vcf"],
+    log:
+        "logs/pangenome/haplotypes.log",
+    shell:
+        "curl -o {output} {params.url} 2> {log}"
+
+
+rule rename_haplotype_conftigs:
```
johanneskoester, line 162:
```suggestion
rule rename_haplotype_contigs:
```

johanneskoester (approved):

FelixMoelder (approved):

PR #282, opened 2024-01-24
feat: render canonical transcript source

FelixMoelder:
In some rare cases multiple transcripts for the same variant show up in the datavzrd report.
This happens in case the `canonical` and `mane_plus_clinical` fields state different transcripts as canonical.
As we can not identify the correct one we added both entries to the datavzrd report in the detail view showing black cells for supporting transcripts.
Additionally, a column for Enseml protein ID has been added.


FelixMoelder:
These changes have already been merged in a previous PR.

PR #283, opened 2024-02-14
feat: add ScanITD candidate calling

dawidkrzeciesa:
Added a new variant caller, ScanITD, to improve the detection of internal tandem duplications.


config/config.yaml
```diff
@@ -85,6 +85,12 @@ calling:
   freebayes:
     activate: true
   # See https://varlociraptor.github.io/docs/calling/#generic-variant-calling
+  ScanITD: 
+    activate: false
+    tmpDIR: "/tmp/ScanITD"  # set directory for temporary files
```
johanneskoester, line 90: This should be hidden from the user. jUST USE THE `temp()` functionality of Snakemake.
```suggestion
```

workflow/rules/candidate_calling.smk
```diff
@@ -68,6 +68,101 @@ rule filter_offtarget_variants:
...
+###############
+### ScanITD ###
+###############
+
+ruleorder: bam_index > split_bam_by_chromosome
+rule split_bam_by_chromosome:
+    input:
+        "results/recal/{sample}.bam",
+    output:
+        bam=temp(config["calling"]["ScanITD"]["tmpDIR"]+"/{sample}_chr{chr}.bam"),
```
johanneskoester, line 80: No need for the tmpDIR thing. Just rely on Snakemake's temp mechanism

workflow/scripts/ScanITD.py
```diff
@@ -0,0 +1,564 @@
+#!/usr/bin/env python3
```
johanneskoester, line 1: I'd rather would want the modifications upstream and us use scanITD via a bioconda package.

workflow/rules/candidate_calling.smk
```diff
@@ -68,6 +68,101 @@ rule filter_offtarget_variants:
...
+    conda:
+        "../envs/bcftools.yaml"
+    params:
+        itd_max_length_bp=config["params"]["ScanITD"]["itd_max_length_bp"]
+    resources:
+        mem_mb=8000,
+    threads: 2
+    shell:
+        "(bcftools reheader -f {input.ref_idx} {input.vcf} | bcftools sort --max-mem {resources.mem_mb}M | "
+        "bcftools view -e 'SVLEN > {params.itd_max_length_bp}' -Ob > {output}) 2> {log}"
```
johanneskoester, line 132: Why do we have this max length parameter? If it is needed, I would rather rely on a vembrane filter for this (i.e. leave it to the user, maybe with an example in the filter section of the config.yaml.

workflow/rules/candidate_calling.smk
```diff
@@ -68,6 +68,101 @@ rule filter_offtarget_variants:
...
+
+
+rule bcftools_gather_ScanITD:
+    input:
+        calls=expand("results/candidate-calls/ScanITD/{{sample}}_chr{chr}.ITD.bcf",chr=[str(i) for i in range(1, 23)] + ["X", "Y", "MT"]),
```
johanneskoester, line 137: The hardcoded chromosomes are dangerous, as they will only work for human. The pipeline is species agnostic though. How long does scanITD take on a WGS sample? Is the splitting really necessary?

workflow/rules/common.smk
```diff
@@ -1057,3 +1058,19 @@ def get_delly_excluded_regions():
         )
     else:
         return []
+
+
+def get_itd_regions(wildcards):
+    group=samples.loc[wildcards.sample]["group"]
+    return  f"results/regions/{group}.expanded_regions.filtered.bed"
```
johanneskoester, line 1065: This can be replaced by
`collect("results/regions/{group}.expanded_regions.filtered.bed", group=lookup(query="sample == '{sample}'", cols="group"))`

johanneskoester: Which can then be done within the rule if it is only needed right there.

workflow/rules/common.smk
```diff
@@ -1057,3 +1058,19 @@ def get_delly_excluded_regions():
...
+
+
+def get_itd_bcfs(wildcards):
+    sample_names = samples.loc[samples["group"] == wildcards.group, "sample_name"].tolist()
+    return [f"results/candidate-calls/ScanITD/{x}.ITD.bcf" for x in sample_names]
+
+
+def get_itd_bcfs_index(wildcards):
+    sample_names = samples.loc[samples["group"] == wildcards.group, "sample_name"].tolist()
+    return [f"results/candidate-calls/ScanITD/{x}.ITD.bcf.csi" for x in sample_names]
```
johanneskoester, line 1075:
```python
def get_itd_bcfs(idx=False):
    def inner(wildcards):
        return collect("results/candidate-calls/ScanITD/{sample}.ITD.bcf{suffix}", sample=sample=lookup(query="group == '{group}'", cols="sample_name"), suffix=".csi" if idx else "")

    return inner
```

johanneskoester (changes requested):

johanneskoester:
Thanks a lot! Unfortunately, in addition to my comments there are also by now some conflicts. Please contact @FelixMoelder in case you need help with resolving them.

FelixMoelder:
> Thanks a lot! Unfortunately, in addition to my comments there are also by now some conflicts. Please contact @FelixMoelder in case you need help with resolving them.

Should be resolved :)

PR #288, opened 2024-03-14
feat: add ability to store variants in a population database (BCF)

johanneskoester:
TODO reuse that DB to annotate variants


workflow/scripts/oncoprint.py
FelixMoelder, line 1: Why are these imports necessary while the remaining code did not change?

workflow/rules/population.smk
```diff
@@ -0,0 +1,44 @@
+if is_activated("population/db"):
```
FelixMoelder, line 1: Does it make any difference to restrict these rules being added to the workflow only if a population-db is activated?

workflow/scripts/update_population_db.py
```diff
@@ -0,0 +1,63 @@
...
+    else:
+        return "not"
+
+
+def copy_to_db(bcf_path):
+    shutil.copyfile(bcf_path, db_path)
+
+
+def update_db(bcf_path, group):
+    sp.run(["bcftools", "index", "-f", db_path])
```
FelixMoelder, line 29: Not happy about this. `pysam` and `bcftools` require an index file which has to be created after each update.

workflow/scripts/update_population_db.py
```diff
@@ -0,0 +1,78 @@
...
+
+    match sample_in_db(group):
+        case "only":
+            copy_to_db(bcf_path)
+            return
+        case "one":
```
FelixMoelder, line 36: Can this be solved differently?
In case the sample already exists it needs to be removed from the population db before merging.
`bcftools merge` does not offer a parameter for omitting samples

workflow/rules/datavzrd.smk
```diff
@@ -1,6 +1,8 @@
 rule split_call_tables:
     input:
-        "results/tables/{group}.{event}.variants.fdr-controlled.tsv",
+        calls="results/tables/{group}.{event}.variants.fdr-controlled.tsv",
+        population_db=get_population_db(use_before_update=True),
```
FelixMoelder, line 4: In case of reruns the current sample will be part of the db. Do we ignore the case or should we perform a cleanup rule first?

FelixMoelder:
@johanneskoester This should be ready to go/review

config/config.yaml
```diff
@@ -157,6 +157,21 @@ remove_duplicates:
...
 
+population:
+  db:
+    # Set to true to continuously update a population database (BCF) with each new sample.
+    activate: false
+    # Path to a BCF file to be used to collect variants across mutliple runs of the pipeline.
+    # The path can be outside of the workdir. Snakemake will update the file with each new
+    # sample that occurs.
+    path: ...
+    alias: tumor
```
johanneskoester, line 168: Add a comment explaining the field

config/config.yaml
```diff
@@ -157,6 +157,21 @@ remove_duplicates:
...
+    fdr: 0.05
```
johanneskoester, line 169: Add a comment explaining the field

workflow/rules/population.smk
```diff
@@ -0,0 +1,50 @@
...
+            alias=lookup(dpath="population/db/alias", within=config),
+            keep_fields="^INFO/SVLEN,INFO/SVTYPE,INFO/MATEID,INFO/END,INFO/CIPOS,INFO/CIEND,^FORMAT/AF",
+        conda:
+            "../envs/varlociraptor.yaml"
+        shell:
+            "varlociraptor filter-calls control-fdr --mode local-smart {input} "
+            "--events {params.events} --fdr {params.fdr} | "
+            "bcftools view --samples {params.alias} | "
+            "bcftools annotate --remove {params.keep_fields} | "
+            "bcftools reheader -s <(echo '{wildcards.group}') | "
```
johanneskoester, line 34: use uncompressed bcf output (`-Ou`) to speed up things here

PR #296, opened 2024-04-11
feat: reduce fgbio memory usage

FelixMoelder:
For large samples `fgbio AnnotateBamWithUmis` fails due to memory issues.
This happens as fgbio loads the fully uncompressed UMI fastq file into memory.
We already tried to fix this in a previous PR (https://github.com/snakemake-workflows/dna-seq-varlociraptor/pull/262) by dynamically assigning more memory based on input file size.
Still this does not suffice in many cases.

To get this fixed fgbio suggests a [best practice](https://github.com/fulcrumgenomics/fgbio/blob/main/docs/best-practice-consensus-pipeline.md#phase-1-fastq---grouped-bam) where raw reads are transformed into unmapped bam files while annotating UMIs at the same time.
In our case this is not feasible as reads need to be adapter trimmed by cutadapt first.
But in case of Qiagen reads the UMI lies in front of the adapter sequence ([source](https://resources.qiagenbioinformatics.com/manuals/biomedicalgenomicsanalysis/2110/index.php?manual=Create_custom_Trim_adapter_list_optional.html)) and therefore can not be annotated while transforming the reads into ubam.

Alternatively, `fgbio AnnotateBamWithUmis` allows to input a queryname sorted bam and fastq file reading the UMIs sequentially from the fastq file, first.
fastq files can be sorted by `fgbio SortFastq`. I intended to directly sort the mapped reads by queryname when running `map_reads` but it showed that queryname sorting differs between `fgbio` and `samtools` receiving incompatible files.
So, it is necessary to reorder the mapped reads by `fgbio SortBam` before annotating UMIs.

PS: Not sure if we treat this as fix or feature

### DIFF
--- a/workflow/rules/common.smk
+++ b/workflow/rules/common.smk
@@ -589,6 +589,18 @@ def get_read_group(wildcards):
     )
 
 
+def get_map_reads_sorting_params(wildcards, order_param=False):
+    match (sample_has_umis(wildcards.sample), order_param):
+        case (True, True):
+            return "queryname"
+        case (True, False):
+            return "fgbio"
+        case (False, True):
+            return "coordinate"
+        case (False, False):
+            return "samtools"
+
+
 def get_mutational_burden_targets():
     mutational_burden_targets = []
     if is_activated("mutational_burden"):
--- a/workflow/rules/common.smk
+++ b/workflow/rules/common.smk
@@ -1113,8 +1113,8 @@ def sample_has_umis(sample):
     return pd.notna(extract_unique_sample_column_value(sample, "umi_read"))
 
 
-def get_umi_read_structure(wildcards):
-    return "-s true -r {}".format(
+def get_annotate_umis_params(wildcards):
+    return "--sorted=true -r {}".format(
         extract_unique_sample_column_value(wildcards.sample, "umi_read_structure")
     )
 
--- a/workflow/rules/common.smk
+++ b/workflow/rules/common.smk
@@ -1091,7 +1091,7 @@ def get_umi_fastq(wildcards):
         return expand(
             "results/untrimmed/{S}_{R}.sorted.fastq.gz",
             S=wildcards.sample,
-            R=["fq1", "fq2"]
+            R=["fq1", "fq2"],
         )
     else:
         return umi_read
--- a/workflow/rules/common.smk
+++ b/workflow/rules/common.smk
@@ -396,7 +396,7 @@ def get_sample_datatype(sample):
 def get_markduplicates_input(wildcards):
     aligner = "star" if get_sample_datatype(wildcards.sample) == "rna" else "bwa"
     if sample_has_umis(wildcards.sample):
-        return "results/mapped/{aligner}/{{sample}}.annotated.bam".format(
+        return "results/mapped/{aligner}/{{sample}}.annotated.sorted.bam".format(
             aligner=aligner
         )
     else:
@@ -1084,12 +1084,12 @@ def get_vembrane_config(wildcards, input):
 def get_umi_fastq(wildcards):
     umi_read = extract_unique_sample_column_value(wildcards.sample, "umi_read")
     if umi_read in ["fq1", "fq2"]:
-        return "results/untrimmed/{S}_{R}.fastq.gz".format(
+        return "results/untrimmed/{S}_{R}.sorted.fastq.gz".format(
             S=wildcards.sample, R=umi_read
         )
     elif umi_read == "both":
         return expand(
-            "results/untrimmed/{S}_{R}.fastq.gz", S=wildcards.sample, R=["fq1", "fq2"]
+            "results/untrimmed/{S}_{R}.sorted.fastq.gz", S=wildcards.sample, R=["fq1", "fq2"]
         )
     else:
         return umi_read
@@ -1100,7 +1100,7 @@ def sample_has_umis(sample):
 
 
 def get_umi_read_structure(wildcards):
-    return "-r {}".format(
+    return "-s true -r {}".format(
         extract_unique_sample_column_value(wildcards.sample, "umi_read_structure")
     )
 
--- a/workflow/rules/common.smk
+++ b/workflow/rules/common.smk
@@ -589,8 +589,8 @@ def get_read_group(wildcards):
     )
 
 
-def get_map_reads_sorting_params(wildcards, order_param=False):
-    match (sample_has_umis(wildcards.sample), order_param):
+def get_map_reads_sorting_params(wildcards, ordering=False):
+    match (sample_has_umis(wildcards.sample), ordering):
         case (True, True):
             return "queryname"
         case (True, False):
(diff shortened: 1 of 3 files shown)


FelixMoelder:
The `bwa mem` wrapper supports sorting by `picard` and `samtools`. We could also add `fgbio` as an option but this would bloat the wrapper environment.

workflow/rules/mapping.smk
```diff
@@ -28,6 +30,19 @@ rule merge_untrimmed_fastqs:
...
+    output:
+        temp("results/untrimmed/{sample}_{read}.sorted.fastq.gz"),
+    conda:
+        "../envs/fgbio.yaml"
+    log:
+        "logs/fgbio/sort_fastq/{sample}_{read}.log",
+    shell:
+        "fgbio SortFastq -i {input} -o {output} 2> {log}"
+
+
```
FelixMoelder, line 45: I would prefer to remove this rule and perform sorting directly when merging fastqs in `merge_untrimmed_fastqs` to reduce IO.
Unfortunately piping does not seem to work. I tried `cat {input} | fgbio SortFastq -i /dev/stdin -o {output}` but fgbio keeps failing. Any suggestion how this could be combined?

FelixMoelder:
> And otherwise, I didn't fully understand the queryname sorting problem. Does `fgbio AnnotateBamWithUmis` not allow for using `samtools sort -n` files? What's the exact issue there?
> 
> And would `samtools collate` be an alternative, as this would be quicker:
> 
> https://www.htslib.org/doc/samtools-collate.html

I tried `samtools sort -n` first. But it turned out that the lexicographic order of samtools is not the same as the order of `fgbio SortFastq` and therefore annotation fails.

FelixMoelder:
> Can we not do the standard best practices and then do the adapter trimming once we have gone back to FASTQ files? So basically do the ubam loop before running cutadapt?

We might could do that but I do not see why it should be a best practice to transform fastq -> ubam -> fastq -> trimming.
Also we would need to write the annotated UMIs from the unmapped bam into the description fields of the fastqs again. And ofter mapping the umis need to be written to the bam record over again. All of this appears to me like a worst practice.

dlaehnemann:
Oh, yeah. I totally agree that there doesn't seem to be an "elegant" solution here. But from looking at everything, I think the least redundant work to do, is if we simply do the best practices and only insert the cutadapt adapter trimming in step `1.2` (after `samtools fastq`, before `bwa mem`):
https://github.com/fulcrumgenomics/fgbio/blob/main/docs/best-practice-consensus-pipeline.md#step-12-ubam---mapped-bam

We might have to make sure that cutadapt does not fully discard reads, even if their sequence is completely trimmed. And I think `samtools fastq` in this setup produces interleaved fastq, so we have to make sure cutadapt works properly in this scenario.

BTW, many thanks for taking care of this -- I know it's a lot of hassle!

FelixMoelder:
I just spend a thought on how the "best practice" actually works.
What they do is to annotated the fastqs in the first step and output a ubam file.
This bam-file is then transformed to fastq once again and mapped against the reference.
The actual annotation is then performed by `fgbio ZipperBams` by inputting the annotated ubam and mapped bam file.
So, I do not see why the transformation of the ubam into a fastq is even necessary as one can just use the initial fastqs, map these and then perform annotation. In our case this would just be a separate step. I just did a quick drawing of the workflow (sorry for my handwriting; boxes are rules).
IMG_0010

In that case we would just skip the ubam -> fastq step and still could trim the initial fastqs.

But, is this any better than the current implementation?
The best practice adds two steps:
- annotating umis (by creating ubams)
- annotating umis to mapped reads (using the annotated ubam)

Our implementation also adds two steps:
- query sorting initial fastqs
- annotating bam by initial fastq

If I should guess I would assume that the best practice is a bit faster (annotation ubam vs query sorting) and probably also required less disk space (bam vs untrimmed fastqs).

FelixMoelder:
I just sketched both options to get an impression what the workflow would look like if we consider that both options would require queryname sorted files.

Best Practice:
Best_Practice

Alternative:
Alternative

The only differences are that the best practice requires an additional step (transformation into ubam) but therefore we could use samtools for queryname sorting as we have only bam files for annotation.
The alternative does not require any file type transformation but as queryname sorting is not supported for fastq files by samtools we would need to do queryname sorting by fgbio for fastqs and during mapping.

dlaehnemann:
Looking at the changes, A) is pretty much what you ended up implementing. So let's stick with that solution for now, I'll review the current state.

FelixMoelder:
I would also advocate for solution A.
Queryname sorting the bam via `fgbio SortBam` is not necessary as a distinct rule as I already updated the `bwa  mem` snakemake-wrapper to support fgbio for sorting in addition to samtools and picard.
The only thing I probably should do is to rename the parameter function for `fgbio AnnotateBamWithUmis` and to replace the short sorting parameter by the long one for comprehensibility.

workflow/rules/common.smk
```diff
@@ -396,7 +396,7 @@ def get_sample_datatype(sample):
 def get_markduplicates_input(wildcards):
     aligner = "star" if get_sample_datatype(wildcards.sample) == "rna" else "bwa"
     if sample_has_umis(wildcards.sample):
-        return "results/mapped/{aligner}/{{sample}}.annotated.bam".format(
+        return "results/mapped/{aligner}/{{sample}}.annotated.sorted.bam".format(
```
dlaehnemann, line 399: Should we maybe make the type of `sorted` clear via the name? For example:

```suggestion
        return "results/mapped/{aligner}/{{sample}}.annotated.fgbio_queryname_sorted.bam".format(
```

dlaehnemann (commented):
Just some minor naming things, and a suggestion to move the `get_annotate_umis_params` function inline to the rule for further readability improvements. I think the semantic helper functions can help make it readable in-line. And the snakemake pipe functionality might solve the piping issue you mentioned.

workflow/rules/common.smk
FelixMoelder: When using the piping functionality we can skip the sorting-suffix as I would write the pipe output the `pipe`-directory as we did for cutadapt.

workflow/rules/mapping.smk
FelixMoelder: I've tried this but fgbio also fails with named pipes and even the `--async-io` flag

workflow/rules/common.smk
```diff
@@ -1100,7 +1114,7 @@ def sample_has_umis(sample):
 
 
 def get_umi_read_structure(wildcards):
-    return "-r {}".format(
+    return "-s true -r {}".format(
         extract_unique_sample_column_value(wildcards.sample, "umi_read_structure")
     )
```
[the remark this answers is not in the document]
FelixMoelder: The first suggestion has been added. The second one does not seem to work in this case as lookups are intended in input and output but also because we are currently throwing an error in case there are multiple entries in the dataframe.

PR #297, opened 2024-04-18
feat: add maftools output

dlaehnemann:


config/config.yaml
```diff
@@ -74,6 +74,11 @@ report:
     # select a sample sheet column for stratification
     by-column: condition
 
+# standardized mutation overview plots with maftools:
+# https://bioconductor.org/packages/release/bioc/vignettes/maftools/inst/doc/maftools.html
+maftools:
+  activate: false
```
johanneskoester, line 80: Do you plan to add all the plots to this PR as well?

dlaehnemann: Yes, exactly. So far, it's the quick fix so everybody can use their own R scripts on the MAF files these initial rules provide. But I'll now also add in rules for the plots. So I'll convert the PR to draft, for now.

dlaehnemann: I think I'll actually opt for reviewing and merging this now, and adding the MAFtools script with plots in a separate pull request.

.test/config-giab/config.yaml
```diff
@@ -50,6 +50,9 @@ mutational_burden:
     - somatic_tumor_low
 
 # printing of variants in a matrix, sorted by recurrence
+maftools:
```
johanneskoester, line 53:
```suggestion
maf:
```
Suggestion: we might switch to a different tool in the future, but the output is MAF format, so this way we make the config independent of the tool. Also, I would name the output folders accordingly, i.e. `results/maf` instead of `results/maftools`.

workflow/rules/maftools.smk
```diff
@@ -0,0 +1,36 @@
...
+        ref=genome,
+    output:
+        maf="results/maftools/{group}.{event}.{calling_type}.fdr-controlled.maf",
+    log:
+        "logs/maftools/{group}.{event}.{calling_type}.fdr-controlled.log",
+    conda:
+        "../envs/vcf2maf.yaml"
+    params:
+        genome_build=config["ref"]["build"],
+        species=config["ref"]["species"],
```
johanneskoester, line 26:
```suggestion
        genome_build=lookup(dpath="ref/build", within=config),
        species=lookup(dpath="ref/species", within=config),
```

workflow/rules/maftools.smk
```diff
@@ -0,0 +1,36 @@
+rule group_bcf_to_vcf:
+    input:
+        "results/final-calls/{group}.{event}.{calling_type}.fdr-controlled.bcf",
+    output:
+        temp("results/maftools/{group}.{event}.{calling_type}.fdr-controlled.vcf"),
+    log:
+        "logs/maftools/{group}.{event}.{calling_type}.fdr-controlled.log",
+    params:
+        extra="",
```
johanneskoester, line 9:
```suggestion
```

workflow/rules/maftools.smk
```diff
@@ -0,0 +1,36 @@
...
+    log:
+        "logs/maftools/{group}.{event}.{calling_type}.fdr-controlled.log",
+    conda:
+        "../envs/vcf2maf.yaml"
+    params:
+        genome_build=config["ref"]["build"],
+        species=config["ref"]["species"],
+    shell:
+        "vcf2maf.pl --inhibit-vep "
+        " --retain-info PROB_ARTIFACT,PROB_ABSENT,PROB_SOMATIC,PROB_GERMLINE_HOM,PROB_GERMLINE_HET "
```
johanneskoester, line 29: It is not guaranteed that users have these events in their scenario. I guess it makes sense to infer that from the input bcf.

workflow/rules/maftools.smk
```diff
@@ -0,0 +1,36 @@
...
+        " --retain-fmt DP,AF "
+        " --ncbi-build {params.genome_build} "
+        " --species {params.species} "
+        " --vcf-tumor-id tumor --vcf-normal-id panel_of_normals "
```
johanneskoester, line 33: Is maf always tumor/normal? Then the config should contain a setting for specifying the proper alias to use for both.

workflow/rules/maftools.smk
```diff
@@ -0,0 +1,36 @@
...
+        " --input-vcf {input.vcf} --output-maf {output.maf} --ref-fasta {input.ref} "
```
johanneskoester, line 34: Does maftools read the input vcf multiple times? If not, I suggest to just use an anonymous pipe here and get rid of the bcftools rule above: `<(bcftools view input.bcf)`.

johanneskoester (changes requested):
Just one thing

.test/config-giab/config.yaml
dlaehnemann: Good idea, done in: https://github.com/snakemake-workflows/dna-seq-varlociraptor/pull/297/commits/d38cced638c7e19875eef4ba0856b7d23950ed43

workflow/rules/maftools.smk
dlaehnemann: Commit is here: https://github.com/snakemake-workflows/dna-seq-varlociraptor/pull/297/commits/cce39c9b4a14118af5adef8647d996971fcfbe6b

workflow/rules/maftools.smk
dlaehnemann: Done.

workflow/rules/maftools.smk
dlaehnemann: Yes, this is now done via params functions. This only took a 3 hour refactoring and then a 5 hour cleanup, but it is definitely better this way. We should seriously think about modularizing this workflow -- and making variant calling and fusion calling separate modules. Having both in one makes it harder to maintain.

workflow/rules/maftools.smk
dlaehnemann: Normal is optional, and this is now handled properly via the config.

workflow/rules/maftools.smk
dlaehnemann: Will this work with a commandline option that expects a path?

.test/config-chm-eval/config.yaml
```diff
@@ -176,6 +176,8 @@ tables:
 gene_coverage:
   min_avg_coverage: 5
 
+
+
```
dlaehnemann, line 180:
```suggestion
```

.test/config-giab/config.yaml
```diff
@@ -56,6 +56,8 @@ mutational_signatures:
     - some_id
 
 # printing of variants in a matrix, sorted by recurrence
+
+
```
dlaehnemann, line 60:
```suggestion
```

.test/config-sra/config.yaml
```diff
@@ -95,6 +95,8 @@ params:
 gene_coverage:
   min_avg_coverage: 8
 
+
+
```
dlaehnemann, line 99:
```suggestion
```

.test/config-target-regions/config_multiple_beds.yaml
```diff
@@ -105,6 +105,8 @@ params:
 gene_coverage:
   min_avg_coverage: 5
 
+
+
```
dlaehnemann, line 109:
```suggestion
```

.test/config_primers/config.yaml
```diff
@@ -98,6 +98,8 @@ params:
 gene_coverage:
   min_avg_coverage: 5
 
+
+
```
dlaehnemann, line 102:
```suggestion
```

workflow/rules/maftools.smk
johanneskoester: Yes, that works. As long as it does not open the path multiple times.

workflow/rules/common.smk
```diff
@@ -1026,115 +1040,318 @@ def get_trimmed_fastqs(wc):
...
+            "INFO['END']",
+        ]
+    )
+    # sort columns, keeping only those in sort_order
+    sorted_columns_dict = {k: columns_dict[k] for k in sort_order if k in columns_dict}
+    join_items = ", ".join
+    return {
+        "expr": join_items(sorted_columns_dict.keys()),
+        "header": join_items(sorted_columns_dict.values()),
+    }
```
johanneskoester, line 1354: 1. Why the sorting?
2. The code is quite redundant. Maybe use a loop over the different args that get_and_sort_multi_entry_fields is called with?

config/config.yaml
```diff
@@ -81,6 +81,17 @@ report:
...
 
+# optional: standardized mutation overview plots with maf:
+# https://bioconductor.org/packages/release/bioc/vignettes/maftools/inst/doc/maftools.html
+maf:
+  activate: false
+  # sample alias from scenario.yaml for tumor sample;
+  # only necessary, if activate: true
+  tumor_alias: "tumor"
+  # optional: sample alias from scenario.yaml for normal sample
+  # normal_alias: "normal"
```
johanneskoester, line 92: Does this always work, also for non cancer samples?

johanneskoester: If so, maybe call the fields more generic, e.g. `primary_alias` and `control_alias`. Let us mayhbe have a quick chat.

johanneskoester (changes requested):

workflow/rules/common.smk
```diff
@@ -1026,115 +1040,318 @@ def get_trimmed_fastqs(wc):
...
+        "CHROM": "chromosome",
+        "POS": "position",
+        "REF": "reference allele",
+        "ALT": "alternative allele",
+        "INFO['END']": "end position",
+        # TODO: remove completely if tests pass without this
+        # END and EVENT were previously included, but they seem to be dropped
+        # in both split-call-tables.py and join_fusion_partner.py, so I'll
+        # simply comment them out here for now
+        # "INFO['EVENT']": "event",
```
johanneskoester, line 1148:
```suggestion
        "INFO['EVENT']": "event",
```

config/config.yaml
johanneskoester:
```suggestion
# optional: standardized mutation overview plots with maf:
# https://bioconductor.org/packages/release/bioc/vignettes/maftools/inst/doc/maftools.html
maf:
  activate: false
  # sample alias from scenario.yaml for tumor sample;
  # only necessary, if activate: true
  primary_alias: "tumor"
  # optional: sample alias from scenario.yaml for normal sample
  # control_alias: "normal"
```

workflow/rules/maftools.smk
dlaehnemann: It seems to be doing some non-standard kind of reading, and the construct `--input-vcf <(bcftools view {input.bcf})` does not work, but throws this:
```
===========================================================
ERROR: Provided --input-vcf is missing or empty: /dev/fd/63
===========================================================
```
A regular `bcftools view <filename>` with the same file in the same conda environment does work. So I'm assuming this is an issue of vcf2maf input reading...

workflow/rules/common.smk
dlaehnemann: 1. I added some comments here: https://github.com/snakemake-workflows/dna-seq-varlociraptor/pull/297/commits/b14d8cfa1411fedb9fc5f3b717b225023341fba5
2. I reduced a bit of redundancy here: https://github.com/snakemake-workflows/dna-seq-varlociraptor/pull/297/commits/735624117f409f99abea8ef731de1d25a928c362

config/config.yaml
dlaehnemann: Good idea. The full refactoring is here:
https://github.com/snakemake-workflows/dna-seq-varlociraptor/pull/297/commits/4b1688fcd576c45da7e86b6797c210bfb245ddfb

PR #308, opened 2024-06-18
feat: mutational signatures

FelixMoelder:
Adding plotting mutational signatures using siglasso.
Mutational signatures are plotted by subsampling over minimum allele frequencies applying a stepsize of 0.1.
Signatures identified in a subsample are weighted higher when being identified in following subsample with higher minimum allele frequency.


config/config.yaml
```diff
@@ -57,6 +57,11 @@ mutational_burden:
     - somatic_tumor_medium
     - somatic_tumor_high
 
+mutational_signatures:
+  activate: true
+  events:
+    - some_id
```
johanneskoester, line 63: Add some explanatory comments. Deactivate by default.

workflow/scripts/preprocess_mutational_variants.py
```diff
@@ -0,0 +1,31 @@
+import sys
```
johanneskoester, line 1: Is this script used anywhere? I can't find a rule that refers to it.

workflow/envs/mutational_context.yaml
```diff
@@ -0,0 +1,8 @@
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - biopython =1.83
+  - pysam =0.22
+  - python >=3.7
```
johanneskoester, line 7: Rather pin the python version to something. Further, instead of 3 envs, just have one for all cases. Maybe call it pystats. The list of tools is not so large that we really need three.

johanneskoester (changes requested):

workflow/envs/mutational_context.yaml
FelixMoelder: Joining the `altair`- and `mutational_context`-env into a single one makes sense. But do we really want to join siglasso as it contains R-packages.

workflow/scripts/preprocess_mutational_variants.py
FelixMoelder: Good catch. This was a leftover from the previous try of implementing SigProfilerAssignment.

config/config.yaml
```diff
@@ -57,6 +57,12 @@ mutational_burden:
     - somatic_tumor_medium
     - somatic_tumor_high
 
+# Plotting of known mutational signatures
```
johanneskoester, line 60:
```suggestion
# Quantify known mutational signatures (human only)
```

config/config.yaml
```diff
@@ -57,6 +57,12 @@ mutational_burden:
...
+mutational_signatures:
+  activate: false
+  events:
```
johanneskoester, line 63:
```suggestion
  events:
    # select events (callsets, defined under calling/fdr-control/events) to evaluate
```

workflow/rules/mutational_signatures.smk
```diff
@@ -0,0 +1,97 @@
...
+        "../envs/pystats.yaml"
+    script:
+        "../scripts/create_mutational_context.py"
+
+
+rule download_cosmic_signatures:
+    output:
+        "resources/cosmic_signatures.txt",
+    params:
+        url=lambda wc: "https://cog.sanger.ac.uk/cosmic-signatures-production/documents/COSMIC_v3.4_SBS_{}.txt".format(
```
johanneskoester, line 21:
```suggestion
        # when updating signature version here, also update workflow/resources/cosmic_signature_desc_v3.4.tsv
        url=lambda wc: "https://cog.sanger.ac.uk/cosmic-signatures-production/documents/COSMIC_v3.4_SBS_{}.txt".format(
```

workflow/rules/mutational_signatures.smk
```diff
@@ -0,0 +1,97 @@
...
+
+rule plot_mutational_signatures:
+    input:
+        signatures="results/mutational_signatures/{group}.{event}.annotated.tsv",
+        counts="results/mutational_signatures/{group}.{event}.counts.tsv",
+    output:
+        report(
+            "results/plots/mutational_signatures/{group}.{event}.svg",
+            category="Mutational Signatures",
+            subcategory="{group}",
```
johanneskoester, line 89:
```suggestion
            subcategory="{event}",
```

workflow/rules/mutational_signatures.smk
```diff
@@ -0,0 +1,97 @@
...
+            labels={"event": "{event}"},
```
johanneskoester, line 90:
```suggestion
            labels={"group": "{group}"},
```

johanneskoester (changes requested):

johanneskoester (approved):

PR #310, opened 2024-06-21
fix: fusion sample selection

FelixMoelder:
When stratification is enabled fusion and variant calling is mixed up as there is no differentiation when selecting groups during stratification.
This has been fixed by passing the calling_type.


johanneskoester (approved):

workflow/rules/fusion_calling.smk
```diff
@@ -39,6 +39,11 @@ use rule arriba from fusion_calling with:
     output:
         fusions="results/arriba/{sample}.fusions.tsv",
         discarded="results/arriba/{sample}.fusions.discarded.tsv",
+    params:
+        genome_build=config["ref"]["build"],
+        default_blacklist=True,
+        default_known_fusions=True,
+        extra="-u -f no_genomic_support,genomic_support,no_coverage,mismatches,homopolymer,low_entropy,duplicates,min_support",
```
dlaehnemann: @FelixMoelder Just came across this change you made to the parameters of arriba, and I was wondering if there was some analysis behind this and if this is documented somewhere?

One change is turning off the internal Arriba duplicate marking with `-u`, here's what the `-u` flag says:
> -u : Arriba performs marking of duplicates internally based on identical mapping coordinates. When this switch is set, internal marking of duplicates is disabled and Arriba assumes that duplicates have been marked by a preceding program.

But I from what I can see in the STAR manual PDF, STAR can only mark duplicates if it is given coordinate sorted BAM input. But we just map the reads from FASTQ with STAR. So I would think that using Arriba's internal duplicate marking would make sense.

But did this somehow negatively affect the sensitivity of Arriba?


And the same question for the filters that were deactivated via the `-f` option:
> -f FILTERS : Comma-/space-separated list of filters to disable. By default all filters are enabled.

Did you test out the sensitivity with those turned on or off? Or did you get them from some recommendation?

PR #314, opened 2024-07-17
fix: use list type for group annotations in datavzrd config

huzuner:
This fixes the type error originating from rendering group annotations in datavzrd config.


dlaehnemann (approved):
LGTM.

PR #317, opened 2024-07-23
ci: Update release-please to v4

fxwiegand:
Just a quick PR updating release-please to v4.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -9,7 +9,7 @@ jobs:
   release-please:
     runs-on: ubuntu-latest
     steps:
-      - uses: GoogleCloudPlatform/release-please-action@v4
+      - uses: googleapis/release-please-action@v4
         id: release
         with:
           release-type: simple
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -9,7 +9,7 @@ jobs:
   release-please:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-please.yml
```diff
@@ -9,7 +9,7 @@ jobs:
   release-please:
     runs-on: ubuntu-latest
     steps:
-      - uses: GoogleCloudPlatform/release-please-action@v3
+      - uses: GoogleCloudPlatform/release-please-action@v4
```
dlaehnemann, line 12: I think we should switch to the newer GitHub Action for this, as the old one is deprecated, see:
https://github.com/google-github-actions/release-please-action?tab=readme-ov-file#deprecated-release-please-action---please-use-httpsgithubcomgoogleapisrelease-please-action

So I'd suggest:
```suggestion
      - uses: googleapis/release-please-action@v4
```

And we might have to further adjust permissions and create a personal access token:
https://github.com/googleapis/release-please-action?tab=readme-ov-file#github-credentials

fxwiegand: Yeah that was in fact the reason why I just wanted to update from v3 to v4. But if we can do the credential stuff sure lets go with the newest one.

dlaehnemann: I think the credentials issue also applies to the deprecated version, as these are GitHub Actions limitations. Currently, we always have to manually override the fact, that no tests are run on the release-please pull requests (and we also have no GUI way to trigger them)...

PR #324, opened 2024-09-02
feat: allow excluding events from the report

huzuner:


workflow/rules/common.smk
```diff
@@ -140,14 +140,37 @@ def get_final_output(wildcards):
 
     for calling_type in calling_types:
         if config["report"]["activate"]:
-            final_output.extend(
-                expand(
-                    "results/datavzrd-report/{batch}.{event}.{calling_type}.fdr-controlled",
-                    batch=get_report_batches(calling_type),
```
johanneskoester, line 146: Is get_report_batches now superfluous?

huzuner: Good point! We brought it back.

johanneskoester (approved):

PR #332, opened 2024-10-17
fix: fix excluding events from the report

huzuner:
Before excluding events from the report did not work as it was thought. All the events that had `False` for `report` generation in the config still had datavzrd reports, because `get_calling_events(calling_type)` always returned all events. This fix ensures that only `True` events have datavzrd reports.

## Summary by CodeRabbit

- **New Features**
	- Improved output file path generation for fusions and variants.
	- Enhanced handling of group samples and aliases with new helper functions.
	- Streamlined logic for generating reports and managing mutational burden targets.
	- New configuration options for reporting within false discovery rate control.

- **Bug Fixes**
	- Corrected event name retrieval in output paths.
	- Fixed indentation issues in configuration files.

- **Refactor**
	- Simplified and clarified the output generation logic for better maintainability.

### DIFF
--- a/workflow/rules/common.smk
+++ b/workflow/rules/common.smk
@@ -150,20 +150,7 @@ def get_final_output(wildcards):
                         expand(
                             "results/datavzrd-report/{batch}.{event}.{calling_type}.fdr-controlled",
                             batch=get_report_batches(calling_type),
-                            event=get_calling_events(calling_type),
-                            calling_type=calling_type,
-                        )
-                    )
-                else:
-                    final_output.extend(
-                        expand(
-                            "results/final-calls/{group}.{event}.{calling_type}.fdr-controlled.bcf",
-                            group=(
-                                variants_groups
-                                if calling_type == "variants"
-                                else fusions_groups
-                            ),
-                            event=get_calling_events(calling_type),
+                            event=event,
                             calling_type=calling_type,
                         )
                     )
--- a/workflow/rules/common.smk
+++ b/workflow/rules/common.smk
@@ -154,6 +154,19 @@ def get_final_output(wildcards):
                             calling_type=calling_type,
                         )
                     )
+                else:
+                    final_output.extend(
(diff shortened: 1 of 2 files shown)


workflow/rules/common.smk
```diff
@@ -150,7 +150,7 @@ def get_final_output(wildcards):
                         expand(
                             "results/datavzrd-report/{batch}.{event}.{calling_type}.fdr-controlled",
                             batch=get_report_batches(calling_type),
-                            event=get_calling_events(calling_type),
+                            event=event,
```
dlaehnemann, line 153: If we remove the use of this utility function here, maybe we can instead use it above, to replace this:
```
for event in config["calling"]["fdr-control"]["events"]:
```
with the function call like this:
```
for event in get_calling_events(calling_type):
```
This ensures, that the calling_type variable from the respective for-loop is also respected, fixing a potential bug.

dlaehnemann (commented):
Very good catch!
I would suggest one small change, namely to move the `get_calling_events(calling_type)` to the `for event in [...]` loop, to keep its functionality in (the correct) place.

huzuner:
Makes sense, just applied the change. I will test this on the workflow but can't do it now as the testing workflow needs a rerun of almost all jobs and I can only test this hopefully sometime next week. But if you think that's not very necessary, you can also merge it now or it can also wait.

dlaehnemann:
Hmm, maybe we could even add the new `report: False` entry to one of the events in one of the simple CI tests in the workflow? This way it would get executed and one could either run that test manually and check the resulting report for the omission of the respective event, or check the artifacts from the CI tests, if there is a way to get to them.

huzuner:
Makes sense, I just added it to one of the tests. Now we can manually check this feature on the tests but not sure how this can be achieved by checking the artifacts from the CI tests.

johanneskoester (approved):

PR #334, opened 2024-10-30
fix: fix short observation plot

FelixMoelder:
Short observation plots were not shown anymore due to a type error. Somehow this error occurred since updating to a newer datavzrd wrapper version.


FelixMoelder:
I will merge this without review as it is a no brainer. :)

PR #344, opened 2024-12-03
feat: support for samples w/ and w/o primers

FelixMoelder:
In some experiments, there are samples with and without primers. However, this case was not previously supported because the workflow always expects a panel containing primers to be provided in the `samples.tsv` file. To address this, the workflow now dynamically checks if a panel is provided for primer trimming. Nevertheless, primer trimming must still be activated in the configuration file.

We should consider whether the primer trimming section in the configuration file is still necessary, or if trimming can be inferred from the presence of a panel in the samplesheet. However, we should retain the primer section for setting a custom library length and specifying the paths to the primer files.

## Summary by CodeRabbit

- **New Features**
	- Introduced a function to check for primer presence in sample data.
	- Enhanced logic for handling sample data and file outputs.
	- Added new requirements for sample and unit sheets in configuration.
	- Enabled primer trimming and mutational burden estimation in configuration files.
	- Activated both `delly` and `freebayes` for variant calling in multiple configurations.

- **Bug Fixes**
	- Improved error handling for sample and unit combinations.

- **Chores**
	- Removed environment specification from the `merge_untrimmed_fastqs` rule.
	- Added execution order dependency between `apply_bqsr` and `bam_index` rules.
	- Updated comments for clarity throughout configuration files.

### DIFF
--- a/workflow/rules/common.smk
+++ b/workflow/rules/common.smk
@@ -1373,7 +1373,6 @@ def sample_has_primers(wildcards):
     if config["primers"]["trimming"].get("primers_fa1"):
         return True
 
-   
     # Check for primers in the sample's panel
     sample_name = wildcards.sample
     if (
--- a/workflow/rules/common.smk
+++ b/workflow/rules/common.smk
@@ -367,11 +367,6 @@ def is_paired_end(sample):
     return all_paired
 
 
-def group_is_paired_end(group):
-    samples = get_group_samples(group)
-    return all([is_paired_end(sample) for sample in samples])
-
-
 def get_map_reads_input(wildcards):
     if is_paired_end(wildcards.sample):
         return [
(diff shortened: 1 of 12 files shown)


workflow/rules/common.smk
```diff
@@ -428,7 +419,7 @@ def get_sample_datatype(sample):
 
 
 def get_markduplicates_input(wildcards):
-    aligner = "star" if get_sample_datatype(wildcards.sample) == "rna" else "bwa"
+    aligner = get_aligner(wildcards)
     if sample_has_umis(wildcards.sample):
         return "results/mapped/{aligner}/{{sample}}.annotated.bam".format(
```
johanneskoester, line 424: F-strings?

johanneskoester (changes requested):
Minor things, nice work! I vote for removing the activate mechanism for primer trimming in the config then.

FelixMoelder:
Just updated the workflow to not rely on primer trimming activation in the config anymore.
The updated config now does not allow placeholder strings for `primers_fa1/fa2` anymore as those would be interpreted as global primers without explicit activation. This would be a breaking change which is why we should treat this PR as a major release.

PR #346, opened 2024-12-12
feat: allow setting of custom alignment properties

FelixMoelder:
This PR introduces the definition of custom alignment properties estimated by varlociraptor.
In case custom alignment properties are enabled in the config file property names can be assigned to samples in the samplesheet via a user-defined column. The corresponding alignment properties will be looked up from a tsv sheet containing property names and paths to json-files. In case some sample does not contain a property name (cell entry is NaN) the alignment properties will be estimate by varlociraptor for that sample. 

## Summary by CodeRabbit

- **Configuration Updates**
	- Added new `custom_alignment_properties` section to multiple configuration files
	- New configuration option allows custom alignment property settings
	- Default setting: feature disabled, column set to "panel"

- **Schema Updates**
	- Updated configuration schema to include new `custom_alignment_properties` object
	- Made new configuration section mandatory

- **Workflow Enhancements**
	- Introduced dynamic alignment properties retrieval
	- Added functions to check feature activation and get alignment properties

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -282,3 +282,12 @@ params:
   freebayes:
     min_alternate_fraction: 0.05 # Reduce for calling variants with lower VAFs
     extra: ""
+
+# If activated preprocessed alignment properties can be applied to each sample individually.
+# paths to the alignment properties json-files need to be set in a tsv-file containing a property name and path.
+# alignment properties will be derived from a customizable column in the sample sheet.
+# if not property name is set for a sample the alignment properties will be estimated best on the samples mapping
+custom_alignment_properties:
+  activate: false
+  column: "panel"
+  tsv: "config_alignment_properties.tsv"
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -284,10 +284,10 @@ params:
     extra: ""
 
 # If activated preprocessed alignment properties can be applied to each sample individually.
-# paths to the alignment properties json-files need to be set in a tsv-file containing a property name and path.
-# alignment properties will be derived from a customizable column in the sample sheet.
-# if not property name is set for a sample the alignment properties will be estimated best on the samples mapping
+# Paths to the alignment properties json-files need to be set in a tsv-file containing a property name and path.
+# Alignment properties will be derived from a customizable column in the sample sheet.
+# If not property name is set for a sample the alignment properties will be estimated best on the samples mapping.
 custom_alignment_properties:
   activate: false
   column: "panel"
-  tsv: "config_alignment_properties.tsv"
+  tsv: "config/alignment_properties.tsv"
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -286,9 +286,9 @@ params:
     extra: ""
 
 # If activated preprocessed alignment properties can be applied to each sample individually.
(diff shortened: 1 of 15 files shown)


workflow/rules/common.smk
```diff
@@ -112,6 +112,29 @@ primer_panels = (
...
+            sep="\t",
+            dtype={"name": str, "path": str},
+            comment="#",
+        )
+        .set_index(["name"], drop=False)
+        .sort_index()
+    )
+    if is_activated("custom_alignment_properties")
+    else None
+)
```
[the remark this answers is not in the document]
johanneskoester: No, this is guarded by the config schema.

workflow/rules/common.smk
```diff
@@ -112,6 +112,29 @@ primer_panels = (
...
+    c = config
+    for entry in xpath.split("/"):
+        c = c.get(entry, {})
+    return bool(c.get("activate", False))
+
+
+custom_alignment_props = (
+    (
+        pd.read_csv(
+            config["custom_alignment_properties"]["tsv"],
```
johanneskoester, line 125: Doesn't this fail if there is no global tsv?

FelixMoelder: Yes, but I would expect the user to define one in case this option has been activated.

workflow/rules/common.smk
```diff
@@ -1591,3 +1607,17 @@ def get_delly_excluded_regions():
...
+        if is_activated("custom_alignment_properties"):
+            alignment_prop_column = config["custom_alignment_properties"]["column"]
+            prop_name = extract_unique_sample_column_value(
+                wildcards.sample, alignment_prop_column
+            )
+            if pd.notna(prop_name):
+                return custom_alignment_props.loc[prop_name, "path"]
+        return "results/alignment-properties/{group}/{sample}.json"
+
+    return inner
```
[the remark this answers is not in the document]
FelixMoelder: I thought it was the preferred way to not pass wildcards as param. How do we device when to use an inner function and when not? In assume in case no element of the function access the wildcards directly we can use an inner function?

config/config.yaml
```diff
@@ -284,3 +284,12 @@ params:
   freebayes:
     min_alternate_fraction: 0.05 # Reduce for calling variants with lower VAFs
     extra: ""
+
+# If activated preprocessed alignment properties can be applied to each sample individually.
+# Paths to the alignment properties json-files need to be set in a tsv-file containing a property name and path.
+# Alignment properties will be derived from a customizable column in the sample sheet.
+# If not property name is set for a sample the alignment properties will be estimated best on the samples mapping.
```
johanneskoester, line 291:
```suggestion
# If activated preprocessed alignment properties can be applied to each sample individually.
# Paths to the alignment properties json files need to be set in a tsv file containing a property name and path.
# Alignment properties names will be taken from a customizable column in the sample sheet.
# If no property name is set for a sample or custom_alignment_properties is deactivated the alignment properties will be estimated estimated on the sample's read alignments.
```

johanneskoester (changes requested):

johanneskoester (approved):

PR #360, opened 2025-02-19
fix: avoid mis-parsing `NA` reference (`REF`) alleles in `split-call-tables.py`

dlaehnemann:
pandas `read_csv` has a bunch of default values it parses into `NaN` automatically, among them `NA`:
https://pandas.pydata.org/pandas-docs/version/2.1/reference/api/pandas.read_csv.html#pandas.read_csv

But in our context, this can be a valid reference allele (`REF`) with an unknown nucleotide in the reference sequence at the first position. So we don't want this parsing. Also, `vembrane table` prints missing values as empty column entries, so the empty string should probably be the only automatic parsing to `NaN`.

### DIFF
--- a/workflow/scripts/split-call-tables.py
+++ b/workflow/scripts/split-call-tables.py
@@ -227,10 +227,13 @@ def select_spliceai_effect(calls):
 calls = pd.read_csv(
     snakemake.input[0],
     sep="\t",
-    #The default pandas `NaN` value `NA` is a valid REF allele that we have seen in the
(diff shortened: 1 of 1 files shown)


dlaehnemann:
@coderabbitai your summary is wrong. This is not mainly about correctly parsing empty values, but about not mis-parsing genuine `NA` entries that can occur in the input, which are non-empty entries. Please remove your summary.

FelixMoelder (approved):
Thanks for digging into this!

PR #382, opened 2025-06-06
feat: add complement events

dlaehnemann:
The use case that motivates this, is to generate a datavzrd report table for all the candidate variants that were filtered out due to their `PROB_ARTIFACT` during FDR control filtering.

The only way that I see to cleanly do this, is to generate two variant sets:

1. A regular variant set that excludes artifacts as is usually intended.
2. A second variant set based on the same varlociraptor events, but with `retain-artifacts: true` set.

We can then take the complement of 1 in 2, to extract those candidate variants that are only found in 2 (with `retain-artifacts: true`).

This functionality might also be interesting for other debugging scenarios or intensive inspection of variant sets, but is probably not common enough to warrant putting it in the main default `config/config.yaml` that comes with every `snakedeploy`ment as a module.

### DIFF
--- a/.test/config-sra/config.yaml
+++ b/.test/config-sra/config.yaml
@@ -47,7 +47,7 @@ calling:
     moderate: "ANN['IMPACT'] == 'MODERATE'"
     no_filter: "True"
   fdr-control:
-    threshold: 0.05
+    threshold: 0.01
     # Denote FDR control mode.
     # The default (local-smart) is recommended because it provides the most realistic and intuitive results.
     # See see https://varlociraptor.github.io/docs/filtering for details.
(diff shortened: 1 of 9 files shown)


dlaehnemann:
Also, the branching off of an older release version is intended, as we need this functionality in a workflow, where we don't want to re-trigger analyses that later changes in the workflow would trigger. But we can update this branch, once we have reviewed everything and before we merge it. But then the user can point to the last commit that works for them.

johanneskoester:
Can't you get the same by defining an ordinary event that selects everything with an artifact probability of least 5%?

dlaehnemann:
Yes, that is a good idea. Basically have the option for varlociraptor to also output everything that it removes during fdr-control, and output one file per non-target varlociraptor (scenario) event, including `absent` and `artifact`.

Then I guess we will not merge this pull request here, but I would like to keep the branch around for our current analysis.

PR #440, opened 2026-03-25
feat: add parameter to make bqsr optional 

BiancaStoecker:
Choose input files according to parameter with or without bqsr. Use sample directly in helper functions instead of passing wildcards.

## Summary by CodeRabbit

* **New Features**
  * Base quality recalibration capability is now available and can be configured via new settings.

* **Refactor**
  * Restructured internal helper functions to improve code maintainability.
  * Reorganized workflow rule directives and configuration blocks for consistency.


johanneskoester:
Thanks a lot! I guess instead of merging this immediately, we could also consider to directly use the new read-alignment workflow once it supports the same functionality?

johanneskoester:
But the preparation is definiely still important as a blueprint for the required changes.

workflow/rules/calling.smk
```diff
@@ -42,8 +42,8 @@ rule varlociraptor_preprocess:
         ref=genome,
         ref_idx=genome_fai,
         candidates=access.multi(get_candidate_calls),
-        bam="results/recal/{sample}.bam",
-        bai="results/recal/{sample}.bai",
+        bam=get_sample_bam,
+        bai=lambda wc: get_sample_bam(wc, bai=True),
```
johanneskoester: Rather do these things like this (throughout this PR):
This also simplifies the helper functions, as they just have to implement the standard extension (.bam in this case, as .bai is just the replacement induced by the subpath function).
```suggestion
        bai=subpath(get_sample_bam, with_suffix=".bai")
```

johanneskoester (changes requested):

workflow/rules/calling.smk
BiancaStoecker: Done

workflow/rules/calling.smk
```diff
@@ -25,8 +25,8 @@ rule varlociraptor_alignment_properties:
     input:
         ref=genome,
         ref_idx=genome_fai,
-        bam="results/recal/{sample}.bam",
-        bai="results/recal/{sample}.bai",
+        bam=get_sample_bam,
+        bai=subpath(get_sample_bam, strip_suffix=".bam", with_suffix=".bai"),
```
johanneskoester, line 29: It should already do the replacement without the strip_suffix

workflow/rules/candidate_calling.smk
```diff
@@ -5,7 +5,7 @@ rule freebayes:
         regions="results/regions/{group}.expanded_regions.filtered.bed",
         # you can have a list of samples here
         alns=access.random(lambda w: get_group_bams(w)),
```
johanneskoester: The lambda here is also superfluous

workflow/rules/mapping.smk
```diff
@@ -115,7 +115,7 @@ rule add_read_group:
     input:
         lambda wc: (
             "results/mapped/vg/{sample}.mate_fixed.bam"
-            if sample_has_primers(wc)
+            if sample_has_primers(wc.sample)
             else "results/mapped/vg/{sample}.reheadered.bam"
         ),
```
johanneskoester: Rather use branch here

workflow/rules/mapping.smk
```diff
@@ -165,7 +165,7 @@ rule annotate_umis:
 
 rule mark_duplicates:
     input:
-        bams=get_markduplicates_input,
+        bams=lambda wc: get_markduplicates_input(wc.sample),
```
johanneskoester: Avoid the lambda. Avoid all lambdas.

johanneskoester (changes requested):

PR #445, opened 2026-05-11
feat: add varlociraptor-inspect linkout and ellipsis to oncoprint headers

christykanto:
Add varlociraptor-inspect linkout and ellipsis to oncoprint headers

## Summary by CodeRabbit

* **New Features**
  * Heatmap headers now include legends with per-annotation dynamic titles.
  * Observation cells render as interactive links that open the external inspection tool with sample observations and per-event probability parameters.

* **Improvements**
  * Removed short-observation columns for a cleaner, streamlined variant/fusion display.
  * Observation column display mode set to normal for consistent presentation.

* **Chores**
  * Removed inline visualization specs and legacy observation parsing helpers.

[Review Change Stack](https://app.coderabbit.ai/change-stack/snakemake-workflows/dna-seq-varlociraptor/pull/445)

PR #450, opened 2026-06-08
fix: Fix datavzrd config for non-coding variants view and hide short observations

fxwiegand:
This pull request introduces several updates to the `variant-calls-template.datavzrd.yaml` configuration by hiding the allele frequency distribution column for the non-coding variants view and the short observations from both. Pretty much a follow up with some fixes for #445.

## Summary by CodeRabbit

* **Chores**
  * Extended template configuration to support per-sample short observations columns (optional, hidden by default).
  * Updated noncoding view to hide allele frequency distribution columns.


FelixMoelder (approved):
Just tried this locally and report looks fine now. Thanks!

PR #56, opened 2020-11-11
Call consensus reads

FelixMoelder:
This PR adds a rule for calling rbt consensus reads.

Currently, only overlapping reads are going to be merged, while duplicates remain separated.

While the current behavior is supposed for panel data we could optionally add an additional rule annotation reads duplicate groups.
In case that rule would be activated consensus read calling would also merge duplicates into a single read.
Which might be useful to improve read quality instead of dropping duplicates.

PR #470, opened 2026-09-17
fix: declare the freebayes config keys the workflow actually reads

Adelagric:
The schema lists `min_allele_frequency` under `params/freebayes`, but the rule has read `min_alternate_fraction` since #56, and `extra` was never declared. So neither key is validated: on `master`, `min_alternate_fraction: "oops"` passes `validate()` and only fails inside freebayes. With this change it is rejected as `'oops' is not of type 'number'`; the unchanged `.test/config-simple` still passes.

## Summary by CodeRabbit

* **Configuration**
  * Renamed the FreeBayes setting `min_allele_frequency` to `min_alternate_fraction`.
  * Added an optional `extra` text setting for FreeBayes configuration.


johanneskoester (approved):
Thanks a lot!

Issue #352, opened 2025-01-22
feat: avoid freebayes runnning forever on high coverage low-complexity regions (and parallelize more smoothly)


dlaehnemann:
We regularly run into situations where `freebayes` does not seem to parallelize particularly well, with some genomic regions taking forever to process, while all other genomic regions have long finished.

I think this usually comes down to the situation that is described in [this issue on the `freebayes` repository](https://github.com/freebayes/freebayes/issues/479). Namely, we are using the [snakemake wrapper for `freebayes`](https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/freebayes.html), which in turn uses the [`fasta_generate_regions.py` script](https://github.com/freebayes/freebayes/blob/master/scripts/fasta_generate_regions.py) to [generate equally sized stretches of reference genome sequence to process in parallel](https://github.com/snakemake/snakemake-wrappers/blob/917ba7b6f2f3c50021a7aabc44e2b117b4f60d84/bio/freebayes/wrapper.py#L48). There are stretches of low complexity genome, in our current case [an only-G repeat in this region on chromosome 5 of the human genome](https://oct2024.archive.ensembl.org/Homo_sapiens/Share/5196750fee44a24326724e6adc9cc2d0?redirect=no) (I think it was this), which can accumulate a lot of coverage, even in targeted sequencing data. This will lead to freebayes on such regions running for a very long time on a single CPU (as in days). And if the snakemake workflow reserved a lot more cores for a freebayes job, say 32, all the other CPUs reserved (31 in this case) will sit idle once all other genomic regions have been done. And if you kill the job to avoid wasting the reserved CPU time, you loose all the work that was done by this freebayes job completely.

Thus, I see three possible things we could address here:

1. **Set a limit on the number of alleles to consider at any site.**
   This is [recommended for performance tuning and can be done via the `--use-best-n-alleles` command line argument](https://github.com/freebayes/freebayes?tab=readme-ov-file#performance-tuning). The default here is `all`, which I guess can easily balloon in low-complexity regions. So I am assuming that even a rather permissive value would be fine here, to just avoid crazy out there numbers. But where exactly this lies, I am unsure -- could be that 20 is good, or 12, or 8, or.... I'll try to get a test case out of our above example and test run times with different values.
2. **Improve the generation of regions to process in parallel, most importantly making them contain an equal amount of sequencing data (as opposed to reference genome stretches of equal length).**
[...]

PR #471, opened 2026-09-17
feat: expose freebayes --use-best-n-alleles via params/freebayes

Adelagric:
Implements the option agreed on in #352.

freebayes evaluates every allele observed at a site by default. On a high coverage low complexity stretch (e.g. a poly-G run at ~30000× in targeted data) this keeps one core busy for days while the other cores reserved by the rule sit idle.

This adds an optional `use_best_n_alleles` key under `params/freebayes`, passed through as `--use-best-n-alleles N` (schema: integer ≥ 1). It is unset by default, so existing configurations keep freebayes' current behaviour; the `config.yaml` comment suggests `20` as a starting point but I did not pick a default, since that needs a benchmark on real data (#352 mentions one).

Checked with a dry run on `.test/config-simple`: unset → freebayes command unchanged; `20` → `--use-best-n-alleles 20` appended; `0` → rejected by the schema.

Note: touches the same schema block as the small fix in #470; whichever merges second needs a trivial rebase.


dlaehnemann:
The parameter lookup is nowadays best done with the snakemake helper function `lookup()`, providing a default for when it is not specified:
https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#the-lookup-function

Adelagric:
Done — switched to `lookup(dpath="params/freebayes/use_best_n_alleles", within=config, default=None)`.

johanneskoester (approved):
Thanks a lot, great addition!

